PR #3, opened 2019-07-22
Engine-scripts: run node environment on worker thread

RomanYarik:
Instead of node environment to run on main node thread, moved it to run on it's own worker thread


RomanYarik:
Re-openning in a new PR

PR #16, opened 2019-08-06
Feature: RUN_OPTIONS flag in feature lifecycle

RomanYarik:
* Increased communication timeout to 2 minutes
* added RUN_OPTIONS key in feature setup handler.
 this allows the features to receive the query params in it's setup function
* added debugWarning flag, which will print warning about messages going through the communication which takes too long


packages/core/src/com/communication.ts
```diff
@@ -38,7 +38,8 @@ export class Communication {
     private rootEnvId: string;
     private rootEnvName: string;
     private idsCounter = new MultiCounter();
-    private readonly callbackTimeout = 5000; // 5 seconds
+    private readonly callbackTimeout = 60_000 * 2; // 2 minutes
+    private readonly warnTimeout = 5_000; // 5 seconds
```
AviVahl: `slowThreshold` on ICommunicationOptions

packages/core/src/communication.feature.ts
```diff
@@ -66,21 +67,43 @@ export default new Feature({
     Universal,
     ({
         config: { host, id, topology, maxLogMessages, loggerSeverity, logToConsole, resolvedContexts },
-        loggerTransports
+        loggerTransports,
+        [RUN_OPTIONS]: runOptinos
```
AviVahl: typo

packages/core/src/communication.feature.ts
```diff
@@ -66,21 +67,43 @@ export default new Feature({
...
     }) => {
         // worker and iframe always get `name` when initialized as Environment.
         // it can be overridden using top level config.
         // main frame might not have that configured, so we use 'main' fallback for it.
         let communication: Communication;
+        const debugWarnings = runOptinos.params.has('debugWarnings');
```
AviVahl: export interface O {
    has(key: string): boolean;
    get(key: string): string | null | undefined;
}

PR #42, opened 2019-09-01
Scripts: Run node environments with inspect mode

RomanYarik:
adding --inspect flag, with, if provided, will run the node environment on different processes, which will cause re-evaluation on the node environment code, and will allow better debugging.


packages/scripts/test/node-environments-manager.unit.ts
```diff
@@ -1,8 +1,10 @@
-import { createDisposables } from '@wixc3/engine-test-kit';
 import chai, { expect } from 'chai';
 import chaiAsPromised from 'chai-as-promised';
 import { join } from 'path';
-import { Application } from '../src';
+
+import { createDisposables } from '@wixc3/engine-test-kit';
+
+import { Application } from '../src/';
```
AviVahl: remove postfix /

PR #50, opened 2019-09-08
added supoort for configurable entities other than config

nadavwix:


packages/core/src/types.ts
```diff
@@ -198,20 +198,26 @@ export type SetupHandler<
...
 export type ContextHandler<
     C,
     EnvFilter extends EnvironmentFilter,
     Deps extends SomeFeature[],
     Filter extends NormalizeEnvironmentFilter<EnvFilter> = NormalizeEnvironmentFilter<EnvFilter>
-> = (runningFeatures: RunningFeatures<Deps, Filter>) => C;
+    > = (runningFeatures: RunningFeatures<Deps, Filter>) => C;
 
-export type PartialFeatureConfig<API> = Partial<MapToPartialType<JustFilter<API, Config<any, any>>>>;
+export interface Configurable<T> {
```
barak007: is this enough to identify Configurable?

examples/file-server/fixtures/run.config.ts
```diff
@@ -5,8 +5,6 @@ import FileServer from '../feature/file-server.feature';
  */
 export default [
     FileServer.use({
-        fileServerConfig: {
-            defaultDirName: process.cwd()
-        }
+
```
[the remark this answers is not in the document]
nadavwix: this config is has been removed from the feature, the new type is stronger and this leftover was an error

packages/core/src/entities/feature.ts
```diff
@@ -192,4 +194,11 @@ export class Feature<
 
         return feature;
     }
+    private identifyApis() {
```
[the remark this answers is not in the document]
nadavwix: thats gonna be the entity's job

packages/core/src/types.ts
nadavwix: seems to work

barak007: I don't sure about this. maybe we should use a symbol or __kind .

PR #51, opened 2019-09-10
Communication: server environments host per socket

RomanYarik:
* making sure communication is disposed on close
* changing server communication to be per socket
* on server, runEngineApp will start upon socket connection


AviVahl:
@RomanYarik what's up with this one? irrelevant? (was in my PR review list)

PR #54, opened 2019-09-26
Com: support reloading iframe hosts

RomanYarik:


packages/core/src/com/communication.ts
```diff
@@ -287,6 +331,36 @@ export class Communication {
...
+    private parseHandlerId(handlerId: string, prelude: string) {
+        const [api, method] = handlerId.slice(prelude.length).split('@');
+        return {
+            api,
+            method,
+            handlerId
+        };
+    }
+
+    private reconnectHandler(instanceId: string, data: ListenMessage['data']) {
```
barak007: fail if one of the handlers has error

PR #57, opened 2019-10-23
added slot with attached config

nadavwix:


packages/core/src/entities/configurable-slot.ts
```diff
@@ -0,0 +1,88 @@
...
+        this.sortItems();
+        return this;
+    }
+
+    private getItemIndex(item: T) {
+        if (!this.sortOrder) {
+            return Number.MAX_SAFE_INTEGER;
+        }
+        const idx = this.sortOrder.findIndex(val => val === item.id);
+        return idx === -1 ? Number.MAX_SAFE_INTEGER : idx;
```
RomanYarik: can be reduced to a single return statement

packages/core/src/entities/configurable-slot.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+    private sortItems() {
+        if (!this.sortOrder) {
+            return;
+        }
+        this.items.sort((item1, item2) => {
+            const idx1 = this.getItemIndex(item1);
+            const idx2 = this.getItemIndex(item2);
+            return idx1 > idx2 ? 1 : -1;
```
RomanYarik: ```if(this.sortOrder) { ... }``` will be a bit more readable I think

PR #69, opened 2019-11-11
Com: service method config can define methods that will not callback

RomanYarik:


packages/core/src/com/communication.ts
```diff
@@ -235,36 +252,47 @@ export class Communication {
...
-        origin = this.rootEnvId
-    ): Promise<unknown> {
+    public callMethod({
+        envId,
+        api,
+        method,
+        args,
+        origin = this.rootEnvId,
+        serviceComConfig = {}
+    }: ICallMethodOptions): Promise<unknown> {
```
barak007: This is a performance junction maybe should go back to regular args and also remove default values.

packages/core/test/node/com.spec.ts
```diff
@@ -95,4 +97,36 @@ describe('Communication', () => {
...
+            { id: 'echoService' },
+            {
+                echo: {
+                    emitOnly: true
+                }
+            }
+        );
+        await proxy.echo('Yo!');
+
+        expect(childCallMethodStub).to.have.not.been.called;
```
barak007: something here needs to be checked it should call callMethod on the child

RomanYarik: Written an explanation

PR #75, opened 2019-11-26
Scripts: optional public path for the engine app

RomanYarik:


packages/scripts/src/cli.ts
```diff
@@ -124,17 +126,18 @@ program
     .option('-p ,--port <port>')
     .action(async (path = process.cwd(), cmd: Record<string, string | undefined>) => {
         const { config: configName, outDir = 'dist', feature: featureName, port: preferredPort } = cmd;
-
+        const publicPath = process.env.WCS_PUBLIC_PATH || '/';
```
AviVahl: once at toplevel

PR #78, opened 2019-12-02
Scripts: serializing config file on build

RomanYarik:


packages/scripts/src/application.ts
```diff
@@ -237,15 +237,23 @@ export class Application {
             configName,
             publicPath: normilizedPublicPath
         });
+        const config: TopLevelConfig = [];
         disposables.add(() => close());
-
+        const topLevelConfigs = configName ? configurations.get(configName) ?? undefined : undefined;
```
AviVahl: `?? undefined`

PR #114, opened 2020-02-12
test-kit: handle tabs/windows opened by tests

eladdo92:
Subscribe to newly open tabs/windows from root page and close them after test.

### DIFF
--- a/packages/test-kit/src/with-feature.ts
+++ b/packages/test-kit/src/with-feature.ts
@@ -160,7 +160,14 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
     afterEach(disposeAfterEach.dispose);
 
     const pages = new Set<puppeteer.Page>();
-    afterEach('close pages', () => Promise.all(Array.from(pages).map(page => page.close())).then(() => pages.clear()));
+    afterEach('close pages', async () => {
+        for (const page of pages) {
+            if (!page.isClosed()) {
+                await page.close();
+            }
+        }
+        pages.clear();
+    });
     afterEach('verify no page errors', () => {
         if (capturedErrors.length) {
             const errorsText = capturedErrors.join('\n');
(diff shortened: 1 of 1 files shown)


packages/test-kit/src/with-feature.ts
```diff
@@ -218,14 +225,22 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
...
+                pages.add(page);
+
+                page.on('pageerror', e => {
+                    capturedErrors.push(e);
+                    console.error(e);
+                });
+
+                // Emitted when the page opens a new tab or window
+                page.on('popup', handlePage);
+            })(page);
```
AviVahl: just call `handlePage(page);` right after the `newPage` call.

packages/test-kit/src/with-feature.ts
```diff
@@ -218,14 +225,22 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
             if (defaultViewport) {
                 await page.setViewport(defaultViewport);
             }
-            pages.add(page);
-            page.on('pageerror', e => {
-                capturedErrors.push(e);
-                console.error(e);
-            });
+
+            (function handlePage(page: puppeteer.Page) {
```
AviVahl: avoid overshadowing parameters. not sure how this is not blowing lint. :/
`subscribeToPage` might be a better name.

packages/test-kit/src/with-feature.ts
```diff
@@ -235,8 +235,10 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
                 });
 
                 // Emitted when the page opens a new tab or window
-                page.on('popup', handlePage);
-            })(page);
+                page.on('popup', trackPage);
+            }
+
+            trackPage(page);
```
AviVahl: call this right after the `newPage` call.

PR #122, opened 2020-03-15
Com: Adding IPC host  to communication

### LINKED ISSUE #120: Implement communication transport over Electron IPC
When running inside electron, we want to start environments that talk to other environments via the built-in IPC.

RomanYarik:
Adding IPC host to enable communication over process protocol.
Also added electron environment type, and handled it in communication.

Closes #120


packages/core-node/src/ipc-host.ts
```diff
@@ -0,0 +1,32 @@
...
+        prcocess.on('message', this.onMessage);
+        prcocess.once('disconnect', () => prcocess.off('message', this.onMessage));
+    }
+
+    private onMessage: (...args: any[]) => void = message => {
+        this.emitMessageHandlers(message);
+    };
+
+    public postMessage(data: any) {
+        if (this.prcocess.send) {
```
barak007: do some thing when there is no send

packages/core/src/com/communication.ts
```diff
@@ -193,26 +206,37 @@ export class Communication {
...
         this.handleReady({ from: instanceId } as ReadyMessage);
+        return this.getConnectedEnvironment(instanceId, host);
+    }
+
+    private getConnectedEnvironment(instanceId: string, host: Target) {
         return {
             id: instanceId,
             onDisconnect: (cb: () => void) => {
-                host.subscribers.listeners.add('disconnect', cb);
+                if (host instanceof WsClientHost) {
```
barak007: why only WsClientHost?

packages/core/src/com/communication.ts
```diff
@@ -193,26 +206,37 @@ export class Communication {
...
 
+        const instanceId = env;
         const url = this.topology[env];
-
         if (!url) {
             throw new Error(`Could not find ${envType} topology for ${env}`);
         }
 
-        const instanceId = env;
+        if (typeof url !== 'string') {
```
barak007: how can it be not string?

packages/core/src/com/communication.ts
```diff
@@ -118,9 +118,22 @@ export class Communication {
         const activeEnvironment = endPoint.environments.find(env => env.env === runtimeEnvironmentName)!;
         activeEnvironment.env = endPoint.env;
 
-        return activeEnvironment.envType === 'node'
-            ? this.connect(activeEnvironment as Environment<string, 'node'>)
-            : this.spawn(activeEnvironment);
+        if (activeEnvironment.envType === 'node') {
```
barak007: maybe here the `activeEnvironment` can provide the "register" function implementation

barak007: so we don't need a specific registerElectronEnvironment here

barak007: by the way why registerElectronEnvironment is related to electron? isn't it register and listen?

barak007: it's very similar to useWorker...

PR #123, opened 2020-03-16
Com: environment should provide initializer

RomanYarik:
Currently, the communication holds the "spawn", "connect", "spawnOrConnect", "manage" etc, to be able to launch environment.

Refactored this so that the environment set's it's own initialization


packages/core-node/test/node-com.unit.ts
```diff
@@ -177,14 +177,14 @@ describe('Node communication', () => {
...
 
-        const { onDisconnect } = await clientCom.connect({
-            env: 'server-host',
-            envType: 'node',
-            endpointType: 'single'
-        });
+        expect(onDisconnect).to.not.eq(undefined);
 
-        onDisconnect(spy);
+        onDisconnect!(spy);
```
barak007: reflect type

packages/core/src/com/initializers/contextual.ts
```diff
@@ -0,0 +1,38 @@
...
+            if (!activeEnvironment) {
+                throw new Error(`${runtimeEnvironmentName} cannot be found in definition of ${env} environment`);
+            }
+            const envInitializer = envInitializers[activeEnvironment.env];
+
+            if (!envInitializer) {
+                throw new Error(`environment initializer is not set for ${activeEnvironment.env}`);
+            }
+
+            activeEnvironment.env = env;
```
barak007: don't change the env object

PR #128, opened 2020-03-29
Scripts: refactor node environment launch

RomanYarik:
refactoring engine-scripts node environment related code, to be more flexible.


packages/scripts/src/application.ts
```diff
@@ -23,7 +23,7 @@ import {
 import { createWebpackConfigs } from './create-webpack-configs';
 import { ForkedProcess } from './forked-process';
 import { NodeEnvironmentsManager, LaunchEnvironmentMode } from './node-environments-manager';
-import { createIPC } from './process-communication';
+import { createServerIPC } from './process-communication';
```
AviVahl: back to IPC?

PR #153, opened 2020-04-23
COM: refactor managed iframe

alisey:


packages/core/src/com/initializers/iframe.ts
```diff
@@ -48,33 +47,57 @@ async function useWindow(com: Communication, host: WindowHost, instanceId: strin
...
+    const handleIframeReload = async () => {
+        await com.envReady(instanceId);
+        com.reconnectHandlers(instanceId);
+    };
+    const existingReloadHandler = iframeReloadHandlers.get(host);
+    if (existingReloadHandler) {
+        host.removeEventListener('load', existingReloadHandler);
+    }
+    iframeReloadHandlers.set(host, handleIframeReload);
+    host.addEventListener('load', handleIframeReload);
```
alisey, line 73: Maybe there's a more elegant way to do this, I wanted to prevent attaching multiple 'load' event listeners if this function was called multiple times on the same iframe.

packages/core/src/com/initializers/iframe.ts
```diff
@@ -34,47 +34,75 @@ export function iframeInitializer({
...
+    if (locationChanged) {
+        com.registerEnv(instanceId, win);
+        await com.envReady(instanceId);
+        const handleIframeReload = () => {
+            com.envReady(instanceId)
+                .then(() => com.reconnectHandlers(instanceId))
+                .catch(reportError);
+        };
+
+        host.addEventListener('load', handleIframeReload);
```
alisey, line 75: But this will add another 'load' handler every time this function is called for the same iframe. Exactly what I talked about here: https://github.com/wixplosives/engine/pull/153#discussion_r413693010

packages/core/src/com/initializers/iframe.ts
```diff
@@ -34,47 +34,87 @@ export function iframeInitializer({
...
-    host.addEventListener('load', () => {
-        com.envReady(instanceId).then(reload).catch(reportError);
-    });
+    const locationChanged = await changeLocation(win, host, instanceId, src);
+    if (locationChanged) {
+        com.registerEnv(instanceId, win);
+        await com.envReady(instanceId);
+        const handleIframeReload = () => {
+            /**
+             * when page is reloaded, it is neccasery to clear the environment 'ready' state in order to be able to wait for the 'ready' event coming from the newly reloaded environment
```
alisey, line 73: Typo neccasery -> necessary

PR #163, opened 2020-04-29
fix: type fixes for ts3.9

AviVahl:
found using the RC. verified as working in downstream project as well


AviVahl:
Ref for possible upstream issue: https://github.com/microsoft/TypeScript/issues/38238

PR #165, opened 2020-04-30
infra: test on Node 14

AviVahl:
- Node 14.1.0 fixed broken puppeteer. lets see if anything else broke.
- drop Node 13.


AviVahl:
Fails because of the Node-14-specific-workaround that was added to puppeteer. lol. reported upstream.

PR #192, opened 2020-05-14
 Test-kit: allow running environment on given fs

RomanYarik:
Exposed runEnvironment method which runs an environment based on a provided fs.

Method allows passing resolved context for a contextual environment, as well as config overrides


examples/file-server/test/file-server.unit.ts
```diff
@@ -0,0 +1,25 @@
+import { runEngineEnvironment } from '@wixc3/engine-test-kit';
+import Feature, { server } from '../feature/file-server.feature';
+import { expect } from 'chai';
+import fs from '@file-services/node';
+
+describe('Processing env test', () => {
+    it('reads directory', async () => {
+        const { dispose, engine } = await runEngineEnvironment({
```
nadavwix, line 8: wrap it with a function that accepts the actual feature, and returns the typed api as well

PR #207, opened 2020-05-26
Prevent race conditions between two useIframe() calls

alisey:
* Cancel previous initialization before starting a new one.
* Clear iframe before init, since it can have old scripts still running even after location change.
* Cancel initialization if iframe's location unexpectedly changed during init.
* When initialization is cancelled or when the iframe is unloaded, remove its environment from COM of the environment that spawned it.
* Environment inside of the iframe will no longer reconnect when `window.location.reload()` is invoked from within the iframe. This functionality was making already complicated code impossible to reason about.


packages/core/src/com/initializers/iframe.ts
```diff
@@ -19,10 +18,6 @@ export function iframeInitializer({
 }: IIframeInitializerOptions): EnvironmentInitializer<{ id: string }> {
     return async (com, { env, endpointType }) => {
         const instanceId = com.getEnvironmentInstanceId(env, endpointType);
-
-        if (!iframeElement) {
-            throw new Error('should provide a host provider function to the current iframe to initialize');
-        }
```
alisey, line 25: This seemed unnecessary, since the iframeElement is a required argument.

AviVahl:
click the "Ready for review" button... :P

alisey:
> click the "Ready for review" button... :P

@AviVahl this isn't safe to merge yet, I need to update how 'Reload Preview' button works in WCS first.

alisey:
Because the bug so much depends on timing - the interval between `useIframe()` calls, the time it takes to load the iframe and initialize the COM feature, the time each async step takes - I don't think we could reliably trigger it using automated tests.

PR #276, opened 2020-07-07
Expose launchHttpServer method

liorer:
Exposing `launchHttpServer` to be used in other projects. Needed for launching a simple server unrelated to any features. Other alternative would be to add express to the other project (which already has the engine as a dependency)


AviVahl:
@liorer please clarify the **why** and what's the use case (a PR's description should contain it).

RomanYarik:
It looks like this method can be exposed as a standalone function, and can be used by the Application class.
@AviVahl @liorer What do you think?

liorer:
@RomanYarik I think it would make sense. The only thing that is needed from the class here is `outputPath`, which could be provided as a function parameter.

liorer:
@RomanYarik moved `launchHttpServer` to a standalone function

packages/scripts/src/launch-http-server.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const noContentHandler: express.RequestHandler = (_req, res) => {
+    res.status(204); // No Content
+    res.end();
+};
+
+export async function launchHttpServer({
+   outputPath,
+   httpServerPort = DEFAULT_PORT,
+}: {
```
RomanYarik, line 17: export to type ILaunchHttpServerOptions

packages/scripts/src/launch-http-server.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const DEFAULT_PORT = 3000;
+
+const noContentHandler: express.RequestHandler = (_req, res) => {
+    res.status(204); // No Content
+    res.end();
+};
+
+export async function launchHttpServer({
+   outputPath,
```
RomanYarik, line 15: maybe don't call it `outputPath` but `staticDirPath` since in this context it's just a path to a directory you statically serve

PR #278, opened 2020-07-08
core: feature flag for finishing the run stage

RomanYarik:
adding an indication for a feature that states if the feature finished it's run stage


RomanYarik:
decided to go with #279

PR #279, opened 2020-07-08
core: make run*() async and waiting

AviVahl:
alternative approach to #278

### DIFF
--- a/packages/core/src/entities/feature.ts
+++ b/packages/core/src/entities/feature.ts
@@ -23,7 +23,7 @@ export class RuntimeFeature<
     API extends EntityRecord = EntityRecord
 > {
     private running = false;
-    private runHandlers = new SetMultiMap<string, () => void>();
+    private runHandlers = new SetMultiMap<string, () => unknown>();
     private disposeHandlers = new SetMultiMap<string, DisposeFunction>();
 
     constructor(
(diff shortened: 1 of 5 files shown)


packages/core/src/entities/feature.ts
```diff
@@ -32,24 +32,26 @@ export class RuntimeFeature<
...
         for (const dep of this.feature.dependencies) {
-            context.runFeature(dep, envName);
+            runPromises.push(context.runFeature(dep, envName));
         }
         const envRunHandlers = this.runHandlers.get(envName) || [];
         for (const handler of envRunHandlers) {
-            handler();
+            runPromises.push(handler());
         }
+        await Promise.all(runPromises)
```
RomanYarik, line 54: maybe better to return this, so that consumer could decide weather to await for that or not

packages/core/src/run-engine-app.ts
```diff
@@ -55,10 +55,8 @@ export async function runEngineApp({
     ]);
     const [runningFeature] = allFeatures;
 
-    const engine = new RuntimeEngine([COM.use({ config: { resolvedContexts, publicPath } }), ...config], options).run(
-        runningFeature,
-        envName
-    );
+    const engine = new RuntimeEngine([COM.use({ config: { resolvedContexts, publicPath } }), ...config], options);
+    await engine.run(runningFeature, envName);
```
RomanYarik, line 59: I am not sure we want to await here.. the await should be done in the higher level, as discussed with @barak007

packages/core/src/entities/feature.ts
AviVahl: it's the same. think about it. the only reason to return is if u want to propagate the promised value (not the case).

packages/core/src/run-engine-app.ts
AviVahl: there is no higher level. runEngineApp is a part of the public API of engine-core.

packages/core/src/runtime-engine.ts
```diff
@@ -31,9 +31,11 @@ export class RuntimeEngine {
         for (const feature of features) {
             this.initFeature(feature, envName);
         }
+        const runPromises: Array<Promise<void>> = [];
         for (const feature of features) {
-            this.runFeature(feature, envName);
+            runPromises.push(this.runFeature(feature, envName));
         }
+        await Promise.all(runPromises);
```
barak007, line 38: why do you need to await here can just return the promise

packages/core/src/runtime-engine.ts
```diff
@@ -45,12 +47,12 @@ export class RuntimeEngine {
...
     }
 
-    public runFeature(feature: Feature, envName: string) {
+    public async runFeature(feature: Feature, envName: string): Promise<void> {
         const featureInstance = this.features.get(feature);
         if (!featureInstance) {
             throw new Error('Could not find running feature: ' + feature.id);
         }
-        featureInstance[RUN](this, envName);
+        await featureInstance[RUN](this, envName);
```
barak007, line 55: also here you can just return

packages/core/src/runtime-engine.ts
AviVahl: didn't want values to leak

packages/core/src/runtime-engine.ts
AviVahl: ditto

AviVahl:
non of them actually wait... they just return a promise which awaits (it's always after the logic).

PR #317, opened 2020-08-04
scripts: deprecate resolveFrom

AviVahl:
will be removed in next major


packages/scripts/src/cli.ts
```diff
@@ -253,10 +253,7 @@ program.parse(process.argv);
 
 function preRequire(pathsToRequire: string[], basePath: string) {
     for (const request of pathsToRequire) {
-        const resolvedRequest = resolveFrom(basePath, request);
-        if (!resolvedRequest) {
-            throw new Error(`cannot resolve required module: "${request}"`);
-        }
+        const resolvedRequest = require.resolve(request, { paths: [basePath] });
```
RomanYarik, line 256: nice

PR #353, opened 2020-08-31
Get playground example to work

AdamYahid:
Allow registration of listeners to listen function in code service


AdamYahid:
@RomanYarik Although this does fix the issue, It seems like an odd case, on one hand I would expect a compilation error in the case, on the other hand I am actually just register a handler, which is application logic and not something I would expect the engine to be opinionated about. I don't really have a smart solution, just thought it might be an interesting discussion

AviVahl (approved):

PR #363, opened 2020-09-03
Adamyahid/feature graph

AdamYahid:


packages/scripts/src/application.ts
```diff
@@ -268,6 +269,28 @@ export class Application {
             });
         });
 
+        if (mode === 'development') {
+            app.get('/render-graph', (req, res) => {
+                const featureName = req.query['feature-name'] as string;
+
+                const visitedFeatures = {} as { [propName: string]: number };
+
+                const links = buildFeatureLinks(features.get(featureName)!.exportedFeature, visitedFeatures, 0);
```
RomanYarik, line 278: what if `features.get(featureName)` is not set?

packages/scripts/src/application.ts
```diff
@@ -268,6 +269,28 @@ export class Application {
...
+
+                const graph = {
+                    nodes: Object.keys(visitedFeatures)
+                        .map((name) => ({ name, id: name, group: visitedFeatures[name] }))
+                        .concat({
+                            name: features.get(featureName)!.exportedFeature.id,
```
RomanYarik, line 284: you already retrieved `features.get(featureName)` earlier. put it in a const and reuse

packages/scripts/src/application.ts
```diff
@@ -268,6 +269,28 @@ export class Application {
...
+                const featureName = req.query['feature-name'] as string;
+
+                const visitedFeatures = {} as { [propName: string]: number };
+
+                const links = buildFeatureLinks(features.get(featureName)!.exportedFeature, visitedFeatures, 0);
+
+                const graph = {
+                    nodes: Object.keys(visitedFeatures)
+                        .map((name) => ({ name, id: name, group: visitedFeatures[name] }))
+                        .concat({
```
RomanYarik, line 283:
```
const { id } = features.get(featureName)!.exportedFeature;
const graph = {
    nodes: {
        ...Object.entries(visitedFeatures).map((name, group) => ({ name, id: name, group })), 
        { id, name: id, group: 0 } 
    }
}
```
also, if name and id are the same value, why do we need them both?

packages/scripts/src/feature-dependency-graph.ts
```diff
@@ -0,0 +1,179 @@
+import type { Feature } from '@wixc3/engine-core/src';
+
+export interface Link {
+    source: string;
+    target: string;
+}
+
+export const buildFeatureLinks = (
+    entry: Feature,
+    visitedFeatures: { [propName: string]: number },
```
RomanYarik, line 10: use Map<string, number> as a type,
because it's safer in line 15 to use `if(`visitedFeatures.has(dep.is))` also a bit clearer

packages/scripts/src/feature-dependency-graph.ts
```diff
@@ -0,0 +1,179 @@
...
+    level: number
+): Array<Link> => {
+    const res = [] as Array<Link>;
+    const deps = entry.dependencies;
+    for (const dep of deps) {
```
RomanYarik, line 15: if you are using only thie id, destruct it - `for (const {id} of deps) { ... }`

packages/scripts/src/feature-dependency-graph.ts
```diff
@@ -0,0 +1,179 @@
...
+    source: string;
+    target: string;
+}
+
+export const buildFeatureLinks = (
+    entry: Feature,
+    visitedFeatures: { [propName: string]: number },
+    level: number
+): Array<Link> => {
+    const res = [] as Array<Link>;
```
RomanYarik, line 13: featureLinks?

packages/scripts/src/feature-dependency-graph.ts
```diff
@@ -0,0 +1,179 @@
+import type { Feature } from '@wixc3/engine-core/src';
+
+export interface Link {
+    source: string;
+    target: string;
+}
+
+export const buildFeatureLinks = (
+    entry: Feature,
```
RomanYarik, line 9: you are using only id and dependencies, you can extract it

packages/scripts/src/feature-dependency-graph.ts
```diff
@@ -0,0 +1,179 @@
...
+    visitedFeatures: { [propName: string]: number },
+    level: number
```
RomanYarik, line 11: hierarchyLevel maybe?

RomanYarik (changes requested):

packages/scripts/src/application.ts
```diff
@@ -268,6 +269,28 @@ export class Application {
             });
         });
 
+        if (mode === 'development') {
+            app.get('/render-graph', (req, res) => {
+                const featureName = req.query['feature-name'] as string;
+
+                const visitedFeatures = {} as { [propName: string]: number };
```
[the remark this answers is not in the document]
RomanYarik: after revisiting the usage, maybe even consider Map<string, number>

PR #426, opened 2020-10-07
Path resolution/runtime params/dashboard feature/auto browser open race

AdamYahid:
1. Add relative path resolution for path
2. Path resolution is case of unknown options
3. Open browser window only after dev server is listening
4. Default feature for dashboard for cleaner url when starting dev server

### DIFF
--- a/packages/engineer/src/cli.ts
+++ b/packages/engineer/src/cli.ts
@@ -78,12 +78,12 @@ program
                 runtimeOptions: parseCliArguments(process.argv.slice(3)),
             });
 
-            await new Promise((resolve) => {
+            const { port } = await new Promise((resolve) => {
                 devServerFeature.serverListeningHandlerSlot.register(resolve);
             });
 
             if (!process.send && featureName && configName && openBrowser === 'true') {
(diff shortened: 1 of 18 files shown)


packages/engineer/src/cli.ts
```diff
@@ -9,7 +9,7 @@
 import { resolve } from 'path';
 import program from 'commander';
 import open from 'open';
-import fs from '@file-services/node';
+import fs, { nodeFs } from '@file-services/node';
```
RomanYarik, line 12: fs === nodeFs...

packages/engineer/src/cli.ts
```diff
@@ -74,11 +74,16 @@ program
                 publicConfigsRoute,
                 autoLaunch,
                 engineerEntry,
-                targetApplicationPath: path,
+                targetApplicationPath: nodeFs.existsSync(path) ? resolve(path) : process.cwd(),
```
RomanYarik, line 77: fs.existsSync

RomanYarik: fs.resolve

RomanYarik (approved):
restore package.json versions

PR #431, opened 2020-10-11
explicitly type dev server feature because of ts infered type with pa…

AdamYahid:
…ths issue

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "15.1.2",
+  "version": "15.1.3-rc.2",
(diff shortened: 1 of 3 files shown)


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "15.1.2",
+  "version": "15.1.3-rc.2",
```
AviVahl, line 2: revert

AviVahl (approved):

AviVahl:
branch is no longer needed. issue resolved in #432

PR #448, opened 2020-10-21
core: load features after engine is running

RomanYarik:
allow loading new features to the running engine, after their dependencies are loaded.


RomanYarik:
going with a different solution

PR #480, opened 2020-11-02
scripts: raise socket ping timeout to ten minutes

PeterShershov:
Allowing our socket to stay connected after a long operation.


AviVahl (commented):
It took me less than 10 mins to install Fedora 33 from scratch.
I see no sensible reason for a process to **get blocked** for 10 mins , no matter what the app is doing.

FYI @barak007

PeterShershov:
We are making this configurable.

PeterShershov:
@AviVahl Some procedures can take a lot of time for less powerful machines hence we don't want to disconnect when a long operation gets processed. 15 seconds was definitely too low anyway.

PR #491, opened 2020-11-08
scripts: pass socket config to node-env-manager

PeterShershov:
Pass socket server configuration to node environment manager for `runEnvironmentInNewServer`

### DIFF
--- a/packages/engineer/src/feature/dev-server.dev-server.env.ts
+++ b/packages/engineer/src/feature/dev-server.dev-server.env.ts
@@ -90,14 +90,18 @@ devServerFeature.setup(
             //Node environment manager, need to add self to the topology, I thing starting the server and the NEM should happen in the setup and not in the run
             // So potential dependencies can rely on them in the topology
             application.setNodeEnvManager(
-                new NodeEnvironmentsManager(socketServer, {
-                    configurations,
-                    features,
-                    defaultRuntimeOptions,
-                    port: actualPort,
-                    inspect,
-                    overrideConfig,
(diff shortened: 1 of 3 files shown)


packages/engineer/src/feature/dev-server.dev-server.env.ts
```diff
@@ -90,14 +90,18 @@ devServerFeature.setup(
...
+                    socketServer,
+                    {
+                        configurations,
+                        features,
+                        defaultRuntimeOptions,
+                        port: actualPort,
+                        inspect,
+                        overrideConfig,
+                    },
+                    { ...socketServerOptions, ...configServerOptions }
```
AviVahl, line 103: shouldn't the destructuring be the other way around? (also in code above)
also, it can be calculated once.

AviVahl (changes requested):

packages/engineer/src/feature/dev-server.dev-server.env.ts
PeterShershov: Don't we want the provided socket options to override the default ones (currently `undefined`)?

AviVahl: `{...(engine.config.js), ...(feature.config.js)}`

AviVahl: so that specific config files will be able to override repo-wide engine config...

AviVahl (approved):

PR #492, opened 2020-11-09
chore(deps): socket.io@3

AviVahl:
adjust code to breaking changes

### DIFF
--- a/packages/scripts/src/launch-http-server.ts
+++ b/packages/scripts/src/launch-http-server.ts
@@ -35,7 +35,7 @@ export async function launchHttpServer({
 
     app.use('/favicon.ico', noContentHandler);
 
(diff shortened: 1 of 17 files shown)


AviVahl:
there's a change in behaviour where default CORS headers seem to be more restricted now. we'll may have to override cors options to have different ports on localhost communicating with each other.

packages/scripts/src/launch-http-server.ts
```diff
@@ -35,7 +35,7 @@ export async function launchHttpServer({
 
     app.use('/favicon.ico', noContentHandler);
 
-    const socketServer = io(httpServer, socketServerOptions);
+    const socketServer = new io.Server(httpServer, { ...socketServerOptions, cors: {} });
```
AdamYahid, line 38: This blocks the user from setting their own cors value, perhaps we should default to {} only of no cors key on socketServerOptions?

PR #493, opened 2020-11-09
(feature) - Pass runtime options to preload - node

AdamYahid:


packages/scripts/src/node-environment.ts
```diff
@@ -39,7 +39,12 @@ export async function runNodeEnvironment({
         );
     }
     const featureLoader = new FeatureLoadersRegistry(new Map(Object.entries(featureLoaders)), resolvedContexts);
-    const loadedFeatures = await featureLoader.getLoadedFeatures(featureName);
+    const optionsRecord: Record<string, string | boolean> = {};
+
+    for (const opt of options || []) {
```
AviVahl, line 44: [key, val]

AviVahl (approved):

PR #504, opened 2020-11-12
feat: allow providing webpack config for building engine application

RomanYarik:
allow to pass path to a webpack config to build the application with

### DIFF
--- a/packages/engineer/src/feature/dev-server.dev-server.env.ts
+++ b/packages/engineer/src/feature/dev-server.dev-server.env.ts
@@ -62,6 +62,7 @@ devServerFeature.setup(
             outputPath,
             featureDiscoveryRoot,
             socketServerOptions,
+            webpackConfigPath,
         } = devServerConfig;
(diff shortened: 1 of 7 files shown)


packages/engineer/src/feature/dev-server.dev-server.env.ts
```diff
@@ -70,8 +71,12 @@ devServerFeature.setup(
 
         run(async () => {
             // Should engine config be part of the dev experience of the engine????
-            const { require, socketServerOptions: configServerOptions, serveStatic } =
-                (await application.getEngineConfig()) ?? {};
+            const {
+                require,
+                socketServerOptions: configServerOptions,
+                serveStatic,
+                webpackConfigPath: definedWebpackConfigPath,
```
AviVahl, line 78: this should not be loaded from engine config.

RomanYarik: why?

PR #507, opened 2020-11-15
test: Add test for context for feature without that context

AdamYahid:


examples/preload/feature/non-contextual.procEnv.nodeCtx.preload.ts
```diff
@@ -0,0 +1,5 @@
+import nonContextualFeature from './non-contextual.feature';
+import { procEnv } from './contextual.feature';
+
+globalThis.envMessages = [...(globalThis.envMessages ?? []), 'procEnvContextualNoContextPreload'];
+nonContextualFeature.setup(procEnv, () => undefined);
```
AviVahl, line 5: setup call in a preload file?

PR #558, opened 2020-12-24
fix(scripts): resolve external features from proper path

RomanYarik:
currently resolving external features base path from the application base path, but the more correct way to do it is if there is an engine config present - from it.


AviVahl:
Don't join the string 'node_nodules'

RomanYarik:
@AviVahl refactored a bit, so now I don't use node_modules string

PR #559, opened 2020-12-24
fix(external-features): properly mapping feature requests

RomanYarik:
The current way we map the external features in webpack is by mapping the feature files, and not the actual requests.


packages/scripts/src/create-webpack-configs.ts
```diff
@@ -1,18 +1,18 @@
 import fs from '@file-services/node';
+import ts from 'typescript';
```
AviVahl, line 2: So scripts now depend on typescript?

barak007 (approved):
Looks fine to me.

PR #567, opened 2020-12-31
feat(core): initialize Worker using a Blob

AviVahl:
allows loading the worker's bundle from another origin


AviVahl:
refactored a bit to simplify

RomanYarik (approved):

PR #572, opened 2021-01-04
fix(core): keep promise-returning fn as-is

### LINKED ISSUE #571: AsyncApi inferring types strangely
Here's a stripped down example: [here](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwgOwM7wMoFEBKA1ASQGEsB9IgeQDkAxAgcTgF44MBPAWwCMIAbAOgBm0ABQByUqSxV6BKmVLZ8xMpVoMxASgDcAKFCRYcGGzDBWwKADcAlgmBEUgmwHMAPABU4oGMGQATVDgAbzgAbSVCEnJqOnoAXQB+AC44AENkNjgAXwA+ZjgPMJMzCEFWXCjVWIZ4vQNoeBLzAFkAV14YGw8MvxgABSgIEDYaNuQELpRPbxBfAKDQiMqVGPUElPTMnIAaOABpWfnAuHQoG2QXfJYcYBg2qGQPU2A3XTgPi2s7BydXT1yYX28V0gLEYCGIzGEymyDEdV0+nAjWMLzgeDSvDawEGEA4NlQrxw1zgOCOfhOuPxhLc4wA1sgIAB3ZD5RKkuCpKkEom5PS6ZpwACCqDYEyFYBsAIKwXenzC-TgFzgdOAbDKhXiXPJCxVao1GEstnsjmQzncHlycs+n3Z7U63V6yAGkNG40mNmmHj2ip8FKCZwuLjg7MVqWQwCslitNptqSK-XiOpOIn4abSUBcqFSF0ElmFmdQmmY+Vz+Zw1tj7NT6cLqSFheLTHyGKxOKG1N5lbjcAjUagemy-IuviggjS9lYeOAEpsIUrqGnLTuAAsIP5UiIm-keHxgBltHAAPRHuAAdRXfjgYDSqEJ-mMEGFovFkr2qDXHQfFLgbTAcC4YAg03bc4G5Gld14fdWV0bJESQNAmhXYDnzFBBZzcDBp1nElZRtE9WE-XgHyZaA6QXJdV3XECS17YAmTAjseRELdaPHXhCU0HZYKAA).

Not sure exactly what is happening here, but something seems to be getting lost in translation.

AviVahl:
closes #571

otherwise, using `ValuePromise<R>` causes `() => boolean` to get re-wrapped as `() => Promise<true> | Promise<false>`

### DIFF
--- a/packages/core/src/com/types.ts
+++ b/packages/core/src/com/types.ts
@@ -51,10 +51,13 @@ export type UnknownFunction = (...args: unknown[]) => unknown;
 export type AsyncApi<T> = {
     [P in keyof T]: P extends keyof ServiceConfig<T>
         ? MultiTanentProxyFunction<T, P extends string ? P : never>
+        : T[P] extends (...args: any[]) => PromiseLike<any>
+        ? T[P]
         : T[P] extends (...args: infer Args) => infer R
-        ? (...args: Args) => ValuePromise<R>
+        ? (...args: Args) => Promise<R>
         : never;
 };
+
 export interface EnvironmentInstanceToken {
     id: string;
 }
--- a/packages/core/test/node/api-type-check.spec.ts
+++ b/packages/core/test/node/api-type-check.spec.ts
@@ -72,6 +72,14 @@ const gui = new Feature({
     },
 });
 
+interface SomeApi {
+    someMethod: () => boolean;
+}
+
+type SomeApiPromisified = AsyncApi<SomeApi>;
+
+typeCheck((_: EQUAL<SomeApiPromisified, { someMethod(): Promise<boolean> }>) => true);
+
 typeCheck(
     (
         _runningDependencies: EQUAL<


packages/core/src/com/types.ts
```diff
@@ -51,10 +51,13 @@ export type UnknownFunction = (...args: unknown[]) => unknown;
 export type AsyncApi<T> = {
     [P in keyof T]: P extends keyof ServiceConfig<T>
         ? MultiTanentProxyFunction<T, P extends string ? P : never>
+        : T[P] extends (...args: any[]) => PromiseLike<any>
+        ? T[P]
         : T[P] extends (...args: infer Args) => infer R
-        ? (...args: Args) => ValuePromise<R>
+        ? (...args: Args) => Promise<R>
```
barak007, line 57: isn't this the only place that uses ValuePromise?

barak007 (approved):
looks fine to me. once a Promise always a Promise I can't tell the difference

packages/core/src/com/types.ts
AviVahl: also used in `multiTenantMethod` in `service-config.ts`. might be a good idea to remove it in the future though.

PR #614, opened 2021-01-31
feat(core): preserve original error message in cross-env API calls

alisey:
Sometimes it's useful to show the error message that occurred in a remote API call to the user:

```ts
try {
    await service.doThing();
} catch (error) {
    setDialogError(error.message);
}
```

But currently we prefix each error with 'Remote call failed with error:', and also include the entire stack trace in the error message, which doesn't look good in the UI.

Additionally, we discard custom error properties, such as error code.

This PR changes error handing to preserve the error name, message, and custom properties. If the error is uncaught, it's useful for debugging to know from which environment it originated, and we add this extra info to `error.stack`. Chrome logs it in full to the console. Unfortunately, Firefox ignores `error.stack` and shows its own stack trace that's not derived from the error object.

### DIFF
--- a/packages/core/src/com/communication.ts
+++ b/packages/core/src/com/communication.ts
@@ -28,7 +28,7 @@ import type {
 
 import { SERVICE_CONFIG } from '../symbols';
 
-import { SetMultiMap, deferred } from '../helpers';
+import { SetMultiMap, deferred, serializeError } from '../helpers';
 import type { Environment, SingleEndpointContextualEnvironment, EnvironmentMode } from '../entities/env';
 import type { IDTag } from '../types';
 import { BaseHost } from './hosts/base-host';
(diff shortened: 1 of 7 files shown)


packages/core/src/com/communication.ts
```diff
@@ -685,8 +685,18 @@ export class Communication {
...
+        if (rec && message.error) {
+            // If the error is not caught later, and logged to the console,
+            // it would be nice to indicate which environment the error came from.
+            // We shouldn't change the error message or the error name because
+            // those could be used by error-handling code. Fortunately, we can
+            // add metadata to the stack trace, and Chrome logs the entire stack
+            // property to the console. Unfortunately, Firefox doesn't.
+            const error = Object.assign(new Error(), message.error);
+            error.stack = REMOTE_CALL_FAILED(message.from, error.stack);
+            rec.reject(error);
```
RomanYarik, line 697: didn't we also discuss the error.type should be ``Remote${error.type}``?

RomanYarik (approved):

packages/core/src/com/communication.ts
alisey: Yes. I've added that information to the stack trace instead, it's still visible in the console when you don't catch the error:

image

This allows us to always preserve the error name while still providing indication that this is a remote call.

PR #639, opened 2021-02-13
refactor(test-kit): browser context for page hook

AviVahl:
rather than hooking to the 'popup' event of every page, we can now add a single listener on the browser context to catch them all.

### DIFF
--- a/packages/test-kit/src/with-feature.ts
+++ b/packages/test-kit/src/with-feature.ts
@@ -205,19 +205,17 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
 
             const browserContext = await browser.newContext(browserContextOptions);
             browserContexts.add(browserContext);
-            const featurePage = await browserContext.newPage();
-            trackPage(featurePage);
-
-            function trackPage(page: playwright.Page) {
-                hookPageConsole(page, isNonReactDevMessage);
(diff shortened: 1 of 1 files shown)


packages/test-kit/src/with-feature.ts
```diff
@@ -205,19 +205,17 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
 
             const browserContext = await browser.newContext(browserContextOptions);
             browserContexts.add(browserContext);
-            const featurePage = await browserContext.newPage();
-            trackPage(featurePage);
-
-            function trackPage(page: playwright.Page) {
-                hookPageConsole(page, isNonReactDevMessage);
+            browserContext.on('page', (page) => {
```
RomanYarik, line 208: no cleanup upon disposal?

RomanYarik (commented):
Is the only benefit here is the fact that we create only one subscriber? or do we gain anything extra?

packages/test-kit/src/with-feature.ts
AviVahl: We're closing the browser context, which closes the pages of that context.

AviVahl:
> Is the only benefit here is the fact that we create only one subscriber? or do we gain anything extra?

Easier to follow/understand...

PR #651, opened 2021-02-22
feat(scripts): allow to resolve local paths on app.run

RomanYarik:
Currently, because how we generate the manifest, we assume that the engine application will run from a context where files can be resolved by referencing them using their package name.
Because of that it references own files using a package requests.
If the entrypoint of the application is running in a context where the package requests cannot be resolved, application.run will fail at evaluation time when running the local feature.

Changing the way we generate the features data in the manifest, in a way where feature definitions will reference to the relative request, and the rest to the package request.

When running application, we will resolve all engine file paths in the manifest to the proper absolute file path

### DIFF
--- a/examples/preload/test/utils.ts
+++ b/examples/preload/test/utils.ts
@@ -24,7 +24,6 @@ export const startServerNewProcess = async ({
         // This is retquired in this set of tests because it validates changes to globals
         nodeEnvironmentsMode: 'forked',
         runtimeOptions,
-        featureDiscoveryRoot: 'dist',
     });
 
     const runningPort = await new Promise<number>((resolve) => {
--- a/examples/preload/test/utils.ts
+++ b/examples/preload/test/utils.ts
@@ -24,6 +24,7 @@ export const startServerNewProcess = async ({
         // This is retquired in this set of tests because it validates changes to globals
         nodeEnvironmentsMode: 'forked',
         runtimeOptions,
+        featureDiscoveryRoot: 'dist',
     });
 
     const runningPort = await new Promise<number>((resolve) => {
--- a/packages/engineer/src/cli-commands.ts
+++ b/packages/engineer/src/cli-commands.ts
@@ -131,7 +131,7 @@ export function buildCommand(program: typeof commander) {
         .option('--external [true|false]', 'build feature as external', parseBoolean, false)
         .option('--eagerEntrypoints [true|false]', 'build feature as external', parseBoolean, false)
         .option(
-            '--libOutDir --featureOutDir <featureOutDir>',
+            '--featureOutDir <featureOutDir>',
             'the directory where the feature library will be published at (relative to the base path). default: "."'
         )
         .option(
--- a/packages/engineer/src/cli-commands.ts
+++ b/packages/engineer/src/cli-commands.ts
@@ -131,8 +131,8 @@ export function buildCommand(program: typeof commander) {
         .option('--external [true|false]', 'build feature as external', parseBoolean, false)
         .option('--eagerEntrypoints [true|false]', 'build feature as external', parseBoolean, false)
         .option(
-            '--featureOutDir <featureOutDir>',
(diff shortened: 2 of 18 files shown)


examples/preload/test/utils.ts
```diff
@@ -24,6 +24,7 @@ export const startServerNewProcess = async ({
         // This is retquired in this set of tests because it validates changes to globals
         nodeEnvironmentsMode: 'forked',
         runtimeOptions,
+        featureDiscoveryRoot: 'dist',
```
AviVahl, line 27: was this needed? afaik the tests of the engine still run from source...? did it not work otherwise?

packages/scripts/src/application.ts
```diff
@@ -383,6 +385,32 @@ export class Application {
         };
     }
 
+    private remapManifestFeaturePaths(manifestFeatures: [string, IFeatureDefinition][]) {
+        const features = new Map<string, IFeatureDefinition>();
+        for (const [featureName, featureDef] of manifestFeatures) {
+            const { filePath, envFilePaths, contextFilePaths, preloadFilePaths } = featureDef;
+            const featureDefinition = { ...featureDef };
+            features.set(featureName, featureDefinition);
```
AviVahl, line 393: this is overwritten on the next line?

packages/scripts/src/application.ts
```diff
@@ -672,6 +754,37 @@ export class Application {
...
+    packageName: string,
+    context: string,
+    filePath: string,
+    isRoot: boolean
+) {
+    const relativeFilePath = fs.relative(context, filePath);
+    const relativeRequest = fs
+        .join(fs.dirname(relativeFilePath), fs.basename(relativeFilePath, fs.extname(relativeFilePath)))
+        .replace(/\\/g, '/');
+    return isRoot
```
AviVahl, line 768: I'm a bit confused. In a mono-repo with:
```
/package.json
/packages/feature-a/package.json
/packages/feature-b/package.json
```
if I'm building `feature-a` into `/packages/feature-a/dist`, the manifest inside that directory should use relative paths for any file inside the `feature-a` package, and package requests for any file outside that package. Is that what's happening here?

Would `isRoot` be true just for `feature-a`? If so, would it because engine build is executed inside `/packages/feature-a` directory?

AviVahl (changes requested):
might be a good time to add tests for this functionality

packages/engineer/src/cli-commands.ts
```diff
@@ -131,8 +131,8 @@ export function buildCommand(program: typeof commander) {
         .option('--external [true|false]', 'build feature as external', parseBoolean, false)
         .option('--eagerEntrypoints [true|false]', 'build feature as external', parseBoolean, false)
         .option(
-            '--featureOutDir <featureOutDir>',
-            'the directory where the published feature file is located (relative to the base path). default: "."'
+            '--libOutDir --featureOutDir <featureOutDir>',
```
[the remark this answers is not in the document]
RomanYarik: reverted the lib outDir
this means the "lib/dist" folder, to where all the sources that will be consumed, will be located at.
for typescript projects this will usually be the outDir

packages/scripts/src/application.ts
RomanYarik: yes and yes.
meaning that if running in monorepo, and running the command from the root, all features found in packages of the monorepo will be isRoot = true. 
I think, that maybe a better approach would be that if featureName was passed, isRoot = true only for the feature being built, otherwise isRoot = false for all.

What do you think?

AviVahl: That depends on what else references `isRoot`, and what assumptions were made there.

There's also the case, where I build `feature-a` (example above) into `/outside-packages/some/elsewhere`, the relative references are no longer the correct approach.

So it's a matter of current context and target directory, and we currently assume `engine build` will:
- be executed within the package you're building.
- output into a folder within that package.

I'm fine with those limitations for now, but anyone trying to go out of the expected assumtions may get into trouble.

PR #653, opened 2021-02-22
chore(deps): stylable@4, webpack@5, latest others

AviVahl:


AviVahl:
passes typechecking and lint.
currently fails executing features built as external features. need to figure out why.

@RomanYarik

packages/scripts/src/create-webpack-configs.ts
```diff
@@ -276,8 +276,10 @@ export function createWebpackConfigForExteranlFeature({
             path: outputPath,
             filename: `[name].${target}.js`,
             chunkFilename: `[name].${target}.js`,
-            libraryTarget: 'var',
-            jsonpFunction: packageName + name,
+            library: {
```
AviVahl, line 279: this change is probably why the tests are failing. need to figure out how this has to be configured on webpack@5.

packages/engineer/src/feature/dev-server.dev-server.env.ts
```diff
@@ -15,27 +15,17 @@ import {
...
-        return {
-            close(cb?: () => void) {
-                if (cb) {
-                    cb();
-                }
-            },
-            invalidate: () => undefined,
-        };
-    };
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
RomanYarik, line 24: why? are the types wrong when importing?

packages/scripts/src/create-webpack-configs.ts
```diff
@@ -276,8 +276,10 @@ export function createWebpackConfigForExteranlFeature({
...
+                type: 'global',
```
RomanYarik, line 280: why did you change to 'global'?

RomanYarik (commented):
no need for single run anymore? how do you make the webpackDevMiddeware to "not watch" for current run?

packages/scripts/src/create-webpack-configs.ts
AviVahl: got a runtime error when using type: `jsonp` with that name. maybe I'm missing something obvious.

packages/engineer/src/feature/dev-server.dev-server.env.ts
AviVahl: known issue. anything that imports `@types/webpack` is currently broken. there are discussions about this in DefinitelyTyped.

packages/scripts/src/create-webpack-configs.ts
RomanYarik: from what I read at https://webpack.js.org/blog/2020-10-10-webpack-5-release/#automatic-unique-naming I see that the `jsonpFunction` is no longer needed.
maybe just restore the `libraryTarget: 'var'` (https://webpack.js.org/configuration/output/#expose-a-variable) and remove the `jsonpFunction`

AviVahl: will try

RomanYarik: @AviVahl seems like removing the libraryTarget whatsoever fixed the issue, and now tests pass

AviVahl:
> no need for single run anymore? how do you make the webpackDevMiddeware to "not watch" for current run?
was just a hack for the days it crashed on mac when watching.

AviVahl:
Thanks for figuring it out. I basically tried everything other than just removing it. XD

AviVahl:
Now, while this passed tests, I'm not really sure it's still compatible with `webpack@4`. I'll need to test this further and make fixes to get it *really* working with both versions

packages/scripts/src/create-webpack-configs.ts
```diff
@@ -291,14 +291,19 @@ export function createWebpackConfigForExternalFeature({
         },
         stats: 'errors-warnings',
     };
+    if (semverLessThan(webpack.version, '5.0.0')) {
```
AviVahl, line 294: this was just added in the last rebase.
will make it compatible with both 4.0.0 and 5.0.0.

PR #657, opened 2021-02-23
fix(scripts): allow re-mapping node requests to be resolved from a given context

RomanYarik:
Currently, if the engine application running from a different context where the external feature is set up, the requests coming from the external feature evaluation to already loaded feature files will not use the already evaluated files, which might cause duplicate registrations.


packages/scripts/src/create-entrypoint.ts
```diff
@@ -401,7 +401,7 @@ function loadExternalFeatures(
     ${addExternalsEventListenerForParentEnvironments()}
     
     ${
-        fetchFeatures
+        fetchFeatures || target === 'electron-renderer'
```
barak007, line 404: why this is not outside?

packages/scripts/src/node-environment.ts
```diff
@@ -71,6 +82,7 @@ export async function runNodeEnvironment({
             }
         }
     }
+    clear();
```
barak007, line 85: why do you clear here?

packages/scripts/src/utils/external-features.ts
```diff
@@ -33,6 +33,29 @@ export function getExternalFeaturesMetadata(
         return {
             name: defaultFeatureName!,
             envEntries,
+            externalRequests: extractDependencyMapping(defaultFeatureName!, features),
         };
     });
 }
+
+export function extractDependencyMapping(
```
barak007, line 41: why this is exported?

barak007 (commented):
In general I want that we go on this PR together after you know the answers to my questions

packages/scripts/src/node-environment.ts
RomanYarik: I thought that after finishing to evaluate the external features, we no longer need to preserve this mapping

packages/scripts/src/utils/external-features.ts
RomanYarik: Because I am reusing this method in other projects

packages/scripts/src/application.ts
```diff
@@ -638,79 +635,10 @@ export class Application {
...
-                            dependencies: dependencies,
-                            exportedEnvs: exportedEnvs,
-                            resolvedContexts: resolvedContexts,
-                            packageName,
-                        } as IFeatureDefinition,
-                    ];
-                }
-            ),
+            features: Array.from(features.entries()).map(([featureName, featureDef]) => {
+                return this.generateReMappedFeature(featureDef, pathToSources, featureName);
```
AdamYahid, line 640: one liner

AdamYahid (approved):

PR #661, opened 2021-02-24
fix(iframe location): fix iframe location change

eifr0980:
Changed the way we change iframe's url to `replace` instead of changing `location.href`, in this way it won't affect the history stack.

Fixes [#3306](https://github.com/wixplosives/component-studio/issues/3306)


alisey (approved):
Nice one! Didn't know that iframes affect the back button.

PR #670, opened 2021-03-01
feat(HtmlWebpackPlugin): added the option to add favicon through config/command

### LINKED ISSUE #168: Favicon support
We want to be able to customize the favicon. Whether we have a path or an actual icon, we are unable to tell the engine to use it.

Regarding Electron, I am not sure how it works but the idea is the same, we want to be able to customize the app's icon.

eifr0980:
# Description 
Added new option to add a favicon to `HtmlWebpackPlugin` using 2 ways:

### Command
Add `--favicon` flag and path to start/build command:
```console
foo@bar:~$ yarn start --favicon ./assets/favicon.ico
```

### Config
Add `favicon` and path to `engine.config.js`:
```js
const path = require('path');
module.exports = {
    favicon: path.join(__dirname, 'favicon.ico'),
};
```

Closes #168


packages/engineer/src/cli-commands.ts
```diff
@@ -53,6 +53,7 @@ export const startCommand: Command = (program) =>
             true
         )
         .option('--title <title>', 'application title to display in browser')
+        .option('--favicon <favicon>', 'application favicon to display in browser')
```
RomanYarik, line 56: I think `faviconPath` will be better here
also about the description, maybe `path to favicon to be displayed in browser environments`

packages/engineer/src/cli-commands.ts
```diff
@@ -83,6 +85,10 @@ export const startCommand: Command = (program) =>
             } = cmd;
 
             try {
+                const basePath = resolve(path);
+                preRequire(pathsToRequire, basePath);
```
RomanYarik, line 89: this already happens inside `startDevServer`. why did you call it here?

packages/engineer/src/feature/gui.dev-server.env.ts
```diff
@@ -14,7 +14,7 @@ guiFeature.setup(
         {
             buildFeature: {
                 engineerWebpackConfigs,
-                devServerConfig: { title, publicConfigsRoute },
+                devServerConfig: { title, favicon, publicConfigsRoute },
```
RomanYarik, line 17: why do we need it here?
this is the configuration for the engineer dashboard

RomanYarik (changes requested):

packages/engineer/src/feature/gui.dev-server.env.ts
eifr0980: added it next to places that has `title`.
So what's the usage of `title` overthere?

RomanYarik: well.. not sure, dont see any good reason.

it's fine by me to keep it like this

PR #675, opened 2021-03-02
fix(scripts): map to requests only if out dir inside base dir scope

RomanYarik:
when mapping own requests to engine files in the manifest, we should not re-map to relative requests, unless the outdir is inside the base path. Otherwise resolution will fail

### DIFF
--- a/packages/scripts/src/application.ts
+++ b/packages/scripts/src/application.ts
@@ -549,6 +549,7 @@ export class Application {
         configName?: string;
         entryPoints: Record<string, Record<string, string>>;
     }) {
+        const outputDirInBasePath = this.outputPath.includes(this.basePath);
         const manifest: IBuildManifest = {
             features: Array.from(features.entries()).map(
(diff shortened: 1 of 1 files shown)


AviVahl (approved):

packages/scripts/src/application.ts
```diff
@@ -549,6 +549,7 @@ export class Application {
         configName?: string;
         entryPoints: Record<string, Record<string, string>>;
     }) {
+        const outputDirInBasePath = this.outputPath.includes(this.basePath);
```
AviVahl, line 552: if anything, startsWith.

AviVahl (changes requested):

packages/scripts/src/application.ts
AviVahl: I'm writing "if anything" cuz this isn't safe on windows due to case insensitive casing.

PR #693, opened 2021-03-15
fix(scripts): properly re-mapping requests in manifest file

RomanYarik:
Using the featureOutDir in in build command, to point to the bundled sources of the application, and fall back to featureDiscovery root


packages/scripts/src/application.ts
```diff
@@ -240,15 +240,18 @@ export class Application {
             })
         );
 
+        const bundledSourcesRoot = fs.resolve(
+            this.basePath,
+            featureOutDir || configFeatureOutDir || this.featureDiscoveryRoot
```
AviVahl, line 245: rename to `sourceRoot`. calculate once, when loading the config file

PR #696, opened 2021-03-18
feat: allow communication between node environment

RomanYarik:
Currently multiple node environments can run, but they can't communicate.

Adding the ability to initiate communication between node environment using either the socket server initializer, or using the new `localNodeEnvironmentInitializer` being exported from the  `@wixc3/engine-core-node` package.

To support the localNodeEnvironmentInitializer, the node environment manager initiates a communication instance, and behaves as a middleman for communication between the node environment.

To support the socketInitializer, the way we load the environments changed a bit.
Now, we first understand the topology of all environments, so we first initiate all http servers, and only after, we launch the environments.
This ensures that all environment will be loaded with the complete topology.


AdamYahid (approved):

AdamYahid:
Looks good to me, but maybe @AviVahl  would have some insights?

packages/core-node/src/local-env-inititializer.ts
```diff
@@ -0,0 +1,19 @@
+import { BaseHost, EnvironmentInitializer } from '@wixc3/engine-core';
+
+export const localNodeEnvironmentInitializer: EnvironmentInitializer<{ id: string }> = async (com, env) => {
```
AviVahl, line 3: why is this async?

packages/core-node/src/local-env-inititializer.ts
```diff
@@ -0,0 +1,19 @@
...
+    if (baseEnvHost instanceof BaseHost) {
+        if (!baseEnvHost.parent) {
+            throw new Error(
+                'Environment was set up with a host without a parent. This initializer is not applicable for this host'
+            );
+        }
+        com.registerEnv(env.env, baseEnvHost.parent);
+    }
+
+    return Promise.resolve({
```
AviVahl, line 16: why Promise wrap?

packages/engineer/src/feature/dev-server.dev-server.env.ts
```diff
@@ -30,7 +30,15 @@ interface WebpackDevMiddleware extends express.Handler {
 
 const attachWSHost = (socketServer: io.Server, envName: string, communication: Communication) => {
     const host = new WsServerHost(socketServer.of(`/${envName}`));
-
+    host.addEventListener('message', ({ data: { from, origin } }) => {
+        // we map both the from and the to, because we change mapping in the host itself, it re-mapps both the origin and the from, for multi-tenancy
+        if (!communication.getEnvironmentHost(from)) {
+            communication.registerEnv(from, host);
```
AviVahl, line 36: on every message we register environment? that seems wrong

packages/scripts/src/application.ts
```diff
@@ -395,7 +395,7 @@ export class Application {
             port,
             router: app,
             nodeEnvironmentManager,
-            close: disposables.dispose,
+            close: () => disposables.dispose(),
```
AviVahl, line 398: why?

packages/scripts/src/launch-http-server.ts
```diff
@@ -39,12 +39,18 @@ export async function launchHttpServer({
...
     return {
         close: async () => {
-            await new Promise((res) => {
+            await new Promise<void>((res, rej) => {
                 for (const connection of openSockets) {
                     connection.destroy();
                 }
                 openSockets.clear();
-                socketServer.close(res);
+                socketServer.close((e) => {
```
AviVahl, line 47: make it 1 liner with trinary

packages/scripts/src/node-environments-manager.ts
```diff
@@ -17,15 +17,29 @@ import {
...
     StartEnvironmentOptions,
     TopLevelConfigProvider,
+    ICommunicationMessage,
 } from './types';
 import type { OverrideConfig } from './config-middleware';
 import { getEnvironmntsForFeature } from './utils/environments';
+import { IPCHost } from '@wixc3/engine-core-node';
 
-type RunningEnvironments = Record<string, number>;
+type RunningEnvironment = {
```
AviVahl, line 26: interface (which, unlike type, can be extended).

and if it's in the public api, directly or indirectly, export it.

packages/scripts/src/ws-environment.ts
```diff
@@ -5,20 +5,36 @@ import { WsServerHost } from '@wixc3/engine-core-node';
...
-            }
+        start: async () => {
+            const wsHost = new WsServerHost(socketServerNamespace);
+
+            const runtimeEngine = await runNodeEnvironment({ ...startEnvironmentOptions, host });
+            const {
+                api: { communication },
+            } = runtimeEngine.engine.getCOM();
+            communication.registerMessageHandler(wsHost);
+            wsHost.addEventListener('message', ({ data: { from, origin } }) => {
```
AviVahl, line 21: I feel like I've commented regarding this code already. is it copy/pasted? is there room for refactoring?

packages/engineer/src/feature/dev-server.dev-server.env.ts
RomanYarik: nope.. on every message to that host we check if it's not registered. it should happen only once

RomanYarik: removed :)

AdamYahid (approved):

PR #752, opened 2021-04-26
feat(engineer): Add support for webpack hot reloading

AdamYahid:
Adds hot reloading for react (only) modules when in development mode, using react-refresh

References
https://github.com/pmmmwh/react-refresh-webpack-plugin/tree/6379b3f6ffcb1be02ea6e6012c8332844d3335df/examples/webpack-hot-middleware


AviVahl:
should be by flag. nothing in the engine forces apps to use React, or even to be a client app for that matter.

AdamYahid:
Yeah there are some API opens here still, but it will be a flag.

packages/scripts/src/create-webpack-configs.ts
```diff
@@ -179,6 +184,20 @@ export function createWebpackConfig({
                 ]
             );
         }
+        if (webpackHot) {
+            // we need the hot middleware client into the env entry
+            entry[envName] = ['webpack-hot-middleware/client', entryPath];
+            plugins.push(new webpack.HotModuleReplacementPlugin());
+            plugins.push(
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+                new ReactRefreshWebpackPlugin({
```
AdamYahid, line 193: @AviVahl @RomanYarik  I'm not really sure about this plugin here
everything I did up to this point is related to webpack hot reloading in general, and is only applicable to applications which have the flag turned on and a web env, so I feel fine about that - along with the fact that we actually are opinionated about webpack

This plugin is only applicable for react applications and should run in conjuncture with the babel/typescript transformation in the users development webpack config, and it feels a bit too opinionated on out part to set it up for them

I would consider documenting something in the lines of
To set up webpack hot reloading for your project you should
1. add ReactRefreshWebpackPlugin to your webpack config
2. Add the babel/typescript transformation to your webpack config
3. engineer start --webpackHot true

What do you think?

AdamYahid: Actually while adding this comment I decided to move on with it, in the next commit I moved the react part of hot reloading to the responsibility of the user - that way the can decide the hot reloading solution, we just add the webpack infra for them

examples/react/webpack.config.hot.js
```diff
@@ -0,0 +1,38 @@
+const { StylableWebpackPlugin } = require('@stylable/webpack-plugin');
```
RomanYarik, line 1: inherit from base one and just extend

packages/scripts/src/create-webpack-configs.ts
```diff
@@ -117,6 +117,8 @@ interface ICreateWebpackConfigOptions {
     overrideConfig?: TopLevelConfig | TopLevelConfigProvider;
     externalFeaturesRoute: string;
     eagerEntrypoint?: boolean;
+    watch?: boolean;
```
RomanYarik, line 120: legacy code?

RomanYarik (approved):

packages/scripts/src/create-webpack-configs.ts
AdamYahid: removed

PR #758, opened 2021-04-29
feat: allow controlling node env config via engineconfig/cli

AdamYahid:


packages/engineer/src/application-proxy-service.ts
```diff
@@ -86,7 +81,8 @@ export class TargetApplication extends Application {
         return super.getFeatureEnvDefinitions(features, configurations);
     }
 
-    public setNodeEnvManager(nem: NodeEnvironmentsManager) {
+    public setNodeEnvManager(nem: NodeEnvironmentsManager, nodeEnvironmentsMode?: LaunchEnvironmentMode) {
```
RomanYarik, line 84: no need for that.
When you do 'new NodeEnvironmentManager(..)' you set up the mode. any specific reason for saving the mode explicitly?

packages/engineer/test/fixtures/engine-config/feature/x.feature.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const NODE_1 = new Environment('node_1', 'node', 'single');
+export const NODE_2 = new Environment('node_2', 'node', 'single');
+export const MAIN = new Environment('main', 'window', 'single');
+
+export default new Feature({
+    id: 'XTestFeature',
+    api: {
+        node_env_1: Service.withType<{ getPid: () => number }>().defineEntity(NODE_1).allowRemoteAccess(),
+        node_env_2: Service.withType<{ getPid: () => number }>().defineEntity(NODE_2).allowRemoteAccess(),
```
RomanYarik, line 11: camelCase

RomanYarik (changes requested):

packages/engineer/src/application-proxy-service.ts
AdamYahid: You pass the mode on the request to runServerEnvironments, not to the node environment manager constructor

RomanYarik: Cool, my bad

PR #760, opened 2021-05-02
hotfix(scripts): fix public path resolution for fetching external features

RomanYarik:
We didn't take into consideration that the publicPath may be an empty string, in which case we don't need to normalize it


AviVahl:
it can also be 'auto' in webpack@5...

RomanYarik:
@AviVahl awesome, I will add this also then.

tnx

RomanYarik:
although the 'auto' is defined at build time, and this is a code running at runtime, so it's a "different" public path

PR #763, opened 2021-05-04
fix(scripts): avoid html webpack plugins overriding each other

RomanYarik:


AviVahl (commented):
fix tests... probably all places that refer to main.html

PR #769, opened 2021-05-06
infra: remove defaulting to webpack-virtual-plugin

### LINKED ISSUE #776: Double webpack bundling
## Expected Behavior
Bundle once?
Notify once?

## Current Behavior
image

## Steps to Reproduce (for bugs)

1. Run some app that uses engine
2. Notice multiple messages on terminal

AdamYahid:
This fixes the issue with an additional build when running engine applications, instead we write into a cache directory and fallback to virtual modules if the directory isn't available for some reason (not a package which isn't really a concern or no write access)

1. There is no cache invalidation since it's not a real cache, we just use it as a store - so we write and read write away, next run we write again.
2. I added the watch because when running unit tests (not end to end, which work fine) for engineer I still saw invalidations, I'm not sure why, but it made sense to me not to watch on files that cannot be edited externally anyway.
3. Perhaps we just need to add the ability to run tests without watch mode, since we don't make changes during test runs

resolves #776


packages/scripts/src/create-webpack-configs.ts
```diff
@@ -164,6 +177,13 @@ export function createWebpackConfig({
             externalFeaturesRoute,
             eagerEntrypoint,
         });
+        if (useVirtualModules) {
```
barak007, line 180: I think we can remove support. and just throw

packages/engineer/src/feature/dev-server.dev-server.env.ts
```diff
@@ -227,7 +227,16 @@ devServerFeature.setup(
                 );
             }
 
-            await Promise.all(compilationPromises);
+            if (compiler.compilers.length > 0) {
```
barak007, line 230: can we use the webpackDevMiddleware api to know when things are done?
also we need to change this entire flow to use one webpackDevMiddleware

AdamYahid: handled in #770

AdamYahid:
After speaking with @barak007 we decided to abandon the generate module loader approach for now since there are several opens there, so we will move on with this implementation.

packages/scripts/src/create-webpack-configs.ts
```diff
@@ -43,27 +44,39 @@ export function createWebpackConfigs(options: ICreateWebpackConfigsOptions): web
...
         const entry: webpack.Entry = {};
         configurations.push(
             createWebpackConfig({
                 ...options,
                 baseConfig,
                 enviroments: webEnvs,
                 target: 'web',
-                virtualModules,
-                plugins,
+                plugins: [],
```
AviVahl, line 77: if we don't really use this field anymore (plugins), remove it from the options of createWebpackConfig

packages/scripts/src/create-webpack-configs.ts
```diff
@@ -43,27 +44,39 @@ export function createWebpackConfigs(options: ICreateWebpackConfigsOptions): web
...
     baseConfig.output.publicPath = publicPath;
+
+    const packageJsonPath = fs.findClosestFileSync(__dirname, 'package.json'); // We want "our" package to extract the package name;
+    if (!packageJsonPath) {
+        // Should never happen
+        throw new Error('Cannot find package.json for determening cache dir');
+    }
+    const packageJson = fs.readJsonFileSync(packageJsonPath) as { name: string };
+
+    const projectCacheDir = findCacheDir({ name: packageJson.name, cwd: context, create: true });
```
AviVahl, line 62: using the package name isn't ideal. it has / in it, as it's a scoped package.

AdamYahid: Besides, what is the problem within being scoped?
It just creates another nested directory, and package names guarantee uniqueness @AviVahl

packages/scripts/src/create-webpack-configs.ts
AdamYahid: Done

packages/scripts/src/create-webpack-configs.ts
```diff
@@ -193,6 +213,7 @@ export function createWebpackConfig({
         ...baseConfig,
         target,
         entry,
+        name: target,
```
AdamYahid, line 216: This isn't directly related to the virtual modules, but it makes the output messages from webpack clearer
https://webpack.js.org/configuration/other-options/#name

packages/scripts/src/application.ts
```diff
@@ -765,6 +766,10 @@ export class Application {
             : fs.findClosestFileSync(basePath, 'webpack.config.js');
         const baseConfig = (typeof baseConfigPath === 'string' ? require(baseConfigPath) : {}) as webpack.Configuration;
 
+        // We must use a path within the same context of the project for proper module resolutions from the entrypoint
+        const modulesPath = fs.resolve(basePath, 'node_modules');
+        const tmpDirPath = nativeFs.mkdtempSync(fs.join(modulesPath, 'engine-entry'), 'utf8');
```
AdamYahid, line 771: I am just writing the entry points into a "temporary module" which is unique because of mktempSync - and I'm cleaning them up when the compiler is done
I'm not a huge fan of generating module directories on the fly, but I needed a way to guarantee uniqueness  and that is safe to clean up when i'm done with
@AviVahl 

As a side node, maybe we should add that API to file-services?

packages/engineer/src/cli-commands.ts
```diff
@@ -115,6 +115,8 @@ export const startCommand: Command = (program) =>
                     nodeEnvironmentsMode,
                 });
 
+                process.on('SIGINT', dispose);
```
AdamYahid, line 118: watchClose is only called on graceful close of webpack compiler, it made sense for me to handle SIGINT in the cli to gracefully cleanup what I can @RomanYarik

AviVahl: I like. works on all 3 platforms?

AdamYahid: https://nodejs.org/api/process.html#process_signal_events

```'SIGINT' from the terminal is supported on all platforms, and can usually be generated with Ctrl+C (though this may be configurable). It is not generated when terminal raw mode is enabled and Ctrl+C is used.```

PR #770, opened 2021-05-06
infra: use multicompiler instead of children

AdamYahid:
Also reordered the dev server file to await for both middleware to be ready, not only the first

### DIFF
--- a/packages/engineer/src/feature/dev-server.dev-server.env.ts
+++ b/packages/engineer/src/feature/dev-server.dev-server.env.ts
@@ -22,11 +22,19 @@ import { Communication, createDisposables } from '@wixc3/engine-core';
 import { buildFeatureLinks } from '../feature-dependency-graph';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const webpackDevMiddleware = require('webpack-dev-middleware') as (compiler: webpack.Compiler) => WebpackMiddleware;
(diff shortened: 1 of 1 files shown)


packages/engineer/src/feature/dev-server.dev-server.env.ts
```diff
@@ -208,27 +216,25 @@ devServerFeature.setup(
...
-                    disposables.add(hotMiddleware.close);
-                    app.use(hotMiddleware);
-                }
                 compilationPromises.push(
                     new Promise<void>((resolve) => {
-                        childCompiler.hooks.done.tap('compiled', () => resolve());
+                        devMiddleware.waitUntilValid(() => resolve());
                     })
                 );
+                if (webpackHot) {
```
barak007, line 231: do you need the none hot one when there is a hotMiddleware ?

AdamYahid: Yes, the hot middleware is responsible for the notification layer, not for the actual compilation management
Se example 
https://github.com/webpack-contrib/webpack-hot-middleware/blob/master/example/server.js#L24

From their docs
This module is only concerned with the mechanisms to connect a browser client to a webpack server & receive updates. It will subscribe to changes from the server and execute those changes using webpack's HMR API.

PR #775, opened 2021-05-09
infra: use our own virtual module loader instead of virtual-modules-plugin

AdamYahid:
Born from an issue with multiple builds in webpack 5

### DIFF
--- a/packages/scripts/src/utils/generate-loader.ts
+++ b/packages/scripts/src/utils/generate-loader.ts
@@ -19,26 +19,26 @@ interface WebpackLoaderContext {
     resourceQuery: string;
 }
 
-interface GenerateFileConfigurationOptions {
-    generate?: (ctx: WebpackLoaderContext) => string;
-}
-
 export default function loader(this: WebpackLoaderContext) {
     this.addContextDependency(this.rootContext);
     this._module.context = this.rootContext;
+
+    // The query comes with the ? - we need to slice it out
     const fileName = this.resourceQuery.slice(1);
     const generatedModule = virtualModules[fileName];
     if (generatedModule === undefined) {
         throw new Error(`No content was generated for virtual module ${this.resourceQuery}`);
     }
+
+    // The virtual module is no longer needed once it's loaded into the module system
+    // No reason to keep on holding it
     delete virtualModules.fileName;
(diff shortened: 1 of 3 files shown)


packages/scripts/src/utils/generate-loader.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export function configLoader(options?: GenerateFileConfigurationOptions) {
+    return {
+        loader: __filename,
+        options,
+        resourceQuery: resourceMatcher,
+    };
+}
+
+export function virtualEntry(filename: string, id = 'entry') {
```
AdamYahid, line 46: I'm not really sure whats going on here, but when accessing ctx.resource, I only get what comes after the !=!, so I needed to add the file name again for it to be in the resourceQuery.

packages/scripts/src/utils/generate-loader.ts
```diff
@@ -0,0 +1,52 @@
+import type webpack from 'webpack';
+
+const resourceMatcher = /^\?generate-loader-(\w+)$/g;
+const virtualModules: Record<string, string> = {};
```
AdamYahid, line 4: I need the "virtual modules" I'm writing to be on some global state, since when the loader calls generate it seems to share the loader instance within the compiler, so I only get the data available to the first time I register the loader. 
Moving the virtualModules here gives me the global state I need

packages/scripts/src/utils/generate-loader.ts
```diff
@@ -0,0 +1,52 @@
...
+
+interface GenerateFileConfigurationOptions {
+    generate?: (ctx: WebpackLoaderContext) => string;
+}
+
+export default function loader(this: WebpackLoaderContext) {
+    this.addContextDependency(this.rootContext);
+    this._module.context = this.rootContext;
+    const fileName = this.resourceQuery.slice(1);
+    const generatedModule = virtualModules[fileName];
```
AdamYahid, line 30: I removed the generate function, since it doesn't provide me with the context I need, I get that from the virtual modules

AdamYahid:
@barak007  This PR isn't ready, and requires a bit more cleanup, but the gist of it is. Would love to hear your comments.

PR #777, opened 2021-05-10
fix(engineer): use "mode" passed to cli

AviVahl:


AdamYahid:
@AviVahl what about passing it from the engine config?

AviVahl:
should be made possible as well... but this was fixed quickly as it actually regressed. the cli used to accept `--mode` and use it. It stopped and we got a report about it not working, thus this PR.

AviVahl:
btw the pattern of destructuring 10+ elements and then regrouping them for the next level is bound to loose flags along the way and cause bugs.
we've already had 5+ PRs just like this one.

PR #779, opened 2021-05-10
fix: set main-dashboard.html as index file for engineer dashboards

AdamYahid:
The issue with not being able to see the project from the root of the monorepo is because we have the dashboard both as the actual dashboard and as a regular feature in the project, so we have 2 middleware serving main-dashboard.html
because the application middleware is registered first it catches the request and serves the one without the default feature being set, so we need to pass it manually.

It can basically happen with every application that serves a main-dashboard environment.

The way to resolve it is to resolve the conflict in the URL which I see several options
1. Namespace the middleware
2. Serve from different servers
3. Assume that our dashboard is the default dashboard and add it as an index - therefor solving the URL conflict.

### DIFF
--- a/packages/engineer/src/feature/dev-server.dev-server.env.ts
+++ b/packages/engineer/src/feature/dev-server.dev-server.env.ts
@@ -23,7 +23,8 @@ import { buildFeatureLinks } from '../feature-dependency-graph';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const webpackDevMiddleware = require('webpack-dev-middleware') as (
-    compiler: webpack.MultiCompiler
+    compiler: webpack.MultiCompiler,
+    options?: { index?: string }
 ) => WebpackDevMiddleware;
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const webpackHotMiddleware = require('webpack-hot-middleware') as (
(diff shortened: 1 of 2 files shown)


RomanYarik:
@AviVahl to remind you - prior to engineer, our dashboard was served from '/'. this really makes sense to me since any other html served, represents an entry point to the application.

packages/engineer/src/feature/dev-server.dev-server.env.ts
```diff
@@ -23,7 +23,8 @@ import { buildFeatureLinks } from '../feature-dependency-graph';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const webpackDevMiddleware = require('webpack-dev-middleware') as (
-    compiler: webpack.MultiCompiler
+    compiler: webpack.MultiCompiler,
+    options?: { index?: string }
```
RomanYarik, line 27: is there any better type for this? this doesn't mean a lot like this.

maybe a more descriptive name, or type from 'webpack-dev-middleware' types

RomanYarik (approved):

packages/engineer/src/feature/dev-server.dev-server.env.ts
AdamYahid: I tried to bring in types from webpack-dev-middleware and webpack-hot-middleware but they have conflicts between those and webpack types

AdamYahid:
@AviVahl perhaps you are right, and maybe we should, but it doesn't change the essence of this PR which is that whatever path we choose might conflict with an env that the user is running in his application. and for our specific case, because we are running engine on the project that defines the dashboard it will always conflict unless we choose a path that is not a valid engine application path - so either namespace things or choose /
I chose to go with / since this is how it was before we stated serving the dashboard as an engine app.

PR #797, opened 2021-05-23
fix: add package.json to entry module

AdamYahid:


AviVahl:
Why?

AdamYahid:
Webpack logs a warning for packages that don't have package.json

AviVahl:
Because it was injected as an actual package instead of a subfolder

PR #798, opened 2021-05-23
fix: gracefully dispose only after engine is fully up

AdamYahid:
only dispose after server set it's server listening handler


AviVahl:
So if I Ctrl+C before it finished bundling it won't clean up?

AdamYahid:
Yes, right now Ari reported that when he Ctrl+C on his machine nothing happens - this in conjuncture with moving to the temp folder should be fine

RomanYarik (approved):

AviVahl:
is us handling ctrl+c still needed with the new virtual modules handling?

PR #822, opened 2021-06-01
infra(scripts): use virtual-modules-plugin

AdamYahid:
revert from writing actual files, so we can avoid cleanup for now

### DIFF
--- a/packages/scripts/src/create-webpack-configs.ts
+++ b/packages/scripts/src/create-webpack-configs.ts
@@ -210,9 +210,6 @@ export function createWebpackConfig({
         },
         plugins: [...basePlugins, ...plugins],
         stats: 'errors-warnings',
-        watchOptions: {
-            ignored: Object.keys(tempEntryModules),
-        },
     };
 }
 
--- a/packages/scripts/src/create-webpack-configs.ts
+++ b/packages/scripts/src/create-webpack-configs.ts
@@ -54,7 +54,7 @@ export function createWebpackConfigs(options: ICreateWebpackConfigsOptions): web
     const configurations: webpack.Configuration[] = [];
 
     if (webEnvs.size) {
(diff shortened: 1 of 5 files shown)


packages/scripts/src/create-webpack-configs.ts
```diff
@@ -212,7 +211,7 @@ export function createWebpackConfig({
         plugins: [...basePlugins, ...plugins],
         stats: 'errors-warnings',
         watchOptions: {
-            ignored: [entryTempDir],
+            ignored: Object.keys(tempEntryModules),
```
AdamYahid, line 214: @AviVahl  This wasn't part of the original code, but seems like ignoring watch on the entry points solves the multiple build issue. and it makes sense not to watch them because they don't actually exist

AviVahl: Ignoring is fine by me, especially if it fixes the multiple builds.

TBH, I didn't know we overrode the 'watchOptions' key. we should merge it with the user's. we actually use it in several places (wcs-electron's `webpack.config.js`), so we want that picked up as well.

packages/scripts/src/create-webpack-configs.ts
```diff
@@ -51,11 +51,10 @@ export function createWebpackConfigs(options: ICreateWebpackConfigsOptions): web
     }
     baseConfig.output.publicPath = publicPath;
 
-    const tempEntryModules: Record<string, string> = {};
-
     const configurations: webpack.Configuration[] = [];
 
     if (webEnvs.size) {
+        const tempEntryModules: Record<string, string> = {};
```
AviVahl, line 57: not temp

AviVahl (changes requested):
we might wanna make sure `watchOptions` is merged with user's. we probably caused issues downstream when we added the temp entries and ignored them

PR #827, opened 2021-06-03
refactor!: remove startEnvironment method from Communication

RomanYarik:
+ create a new initializer that allows starting the iframe environments delayed


AviVahl:
I'm not sure what "delayed" means for the rest of the abstractions. what happens if I'm in another env calling a remote service from that iframe (before it's been initialized)?

nadav-ab:
> I'm not sure what "delayed" means for the rest of the abstractions. what happens if I'm in another env calling a remote service from that iframe (before it's been initialized)?

For services i agree we should add a test.

the need is very apparent in component studio, where a lot of components are dependent on the previewId to talk to it,
but are rendered first time before the iframe is initialized.
now they get a 2nd render when it is, (so they can request data with the id)
and a third time when the data is recieved

RomanYarik:
@nadav-ab @AviVahl I changed the test to ensure that services being called before iframe initialization, are being resolved once iframe is initialized

packages/core/src/com/initializers/contextual.ts
```diff
@@ -3,36 +3,40 @@ import type { SingleEndpointContextualEnvironment, Environment } from '../../ent
 import type { Communication } from '../communication';
 import type { MapBy } from '../../types';
 
-export type EnvironmentInitializers<ENVS extends Environment[]> = {
-    [K in keyof MapBy<ENVS, 'env'>]: EnvironmentInitializer<any>;
+export type EnvironmentInitializers<
+    ENVS extends Environment[],
+    EnvToken extends { id: string } | Promise<{ id: string }>
```
AviVahl, line 8: | Promise can be removed?

packages/core/src/com/initializers/contextual.ts
```diff
@@ -3,36 +3,40 @@ import type { SingleEndpointContextualEnvironment, Environment } from '../../ent
...
-            }
+    if (!activeEnvironment) {
+        throw new Error(`${runtimeEnvironmentName} cannot be found in definition of ${env} environment`);
+    }
 
-            return communication.startEnvironment({ ...activeEnvironment, env }, envInitializer);
-        } else {
-            throw new Error('error');
+    if (activeEnvironment.env in envInitializers) {
+        const key: keyof typeof envInitializers = activeEnvironment.env;
```
AviVahl, line 33: not needed?

packages/core/src/com/initializers/iframe.ts
```diff
@@ -10,47 +11,88 @@ export interface IIframeInitializerOptions {
...
+                  })
+                : startIframe({
+                      ...baseStartIframeParams,
+                      host: iframeElement,
+                  });
+        },
+    };
+}
+
+interface StartIframeBaseParams {
```
AviVahl, line 61: export (also below)

PR #833, opened 2021-06-13
feat(test-kit): support playwright tracing

RomanYarik:
+ make a bit stricter types for getLoadedFeature


packages/test-kit/test/trace.unit.ts
```diff
@@ -0,0 +1,60 @@
+import { expect } from 'chai';
+import { createMemoryFs } from '@file-services/memory';
+import { ensureTracePath, TRACE_FILE_EXT } from '../src/utils';
```
AviVahl, line 3: Use public package name

PR #834, opened 2021-06-15
refactor: better virtual-modules with !=!

AviVahl:
webpack-virtual-modules doesn't work well with webpack@5.
replaced it with custom loader solution that uses !=! to represent new virtual entry module, getting their contents from options.

### DIFF
--- a/packages/scripts/src/virtual-modules-loader.ts
+++ b/packages/scripts/src/virtual-modules-loader.ts
@@ -1,0 +1,39 @@
+import type webpack from 'webpack';
+
+const hasOwnProperty = Object.prototype.hasOwnProperty;
+
+export interface VirtualModulesLoaderOptions {
+    [fileName: string]: string;
+}
+
+const virtualModulesLoader: webpack.LoaderDefinition = function () {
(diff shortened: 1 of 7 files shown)


AviVahl:
FYI @eifr0980

AviVahl:
@barak007

packages/scripts/src/virtual-modules-loader.ts
```diff
@@ -0,0 +1,39 @@
...
+    const options = this.getOptions() as VirtualModulesLoaderOptions;
+    const match = /^\?id=(.*)$/.exec(this.resourceQuery);
+    if (!match) {
+        throw new Error(`invalid resourceQuery: ${this.resourceQuery}`);
+    }
+    const fileName = match[1]!;
+    if (!options[fileName]) {
+        throw new Error(`missing generator for: ${fileName}`);
+    }
+    return options[fileName]!;
```
barak007, line 19: I think a function would allow greater control but as long as the text is working I don't mind.

barak007 (approved):

AdamYahid (approved):

PR #841, opened 2021-06-17
test-kit(feat): add better tracing capabilities

RomanYarik:
Allow providing TRACING environment variable, when where it's passed, it will automatically save a trace for all running tests.

Also, as the value for TRACING it is possible to provide the name of the trace that will be saved at `process.cwd()`.

Allow providing specifying tracing in the suite level.

Both these approaches will use the name of the test as the zip name

### DIFF
--- a/packages/test-kit/src/with-feature.ts
+++ b/packages/test-kit/src/with-feature.ts
@@ -178,6 +178,8 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
         }
     });
 
+    afterEach(disposeAfterEach.dispose);
+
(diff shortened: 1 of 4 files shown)


RomanYarik:
@Ari-Kishon @yardenporat @JoMarton FYI

packages/test-kit/src/with-feature.ts
```diff
@@ -189,6 +201,8 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
         }
     });
 
+    const defaultTracing = typeof suiteTracing === 'boolean' ? {} : suiteTracing;
```
AviVahl, line 204: use same pattern for other parameters of getLoadedFeaturew

PR #901, opened 2021-08-11
fix(scripts): always create external-features.json on build

RomanYarik:


RomanYarik:
because we create an entrypoint which always fetches for the /external-features.json, and when running a built feature, from a context which doesn't serve this path, the evaluation of the entrypoint will fail.

we cannot decide at build time whether we should fetch or not, since we don't know how the user will load external features or not at runtime.

PR #908, opened 2021-08-23
fix(com): fix iframe initializer handling of src parameter

tomermes:
The previous version never used the src that was passed to it, only as a condition. See the compiled version:
image

### DIFF
--- a/packages/core/src/com/initializers/iframe.ts
+++ b/packages/core/src/com/initializers/iframe.ts
@@ -11,7 +11,7 @@ export interface IIframeInitializerOptions {
     managed?: boolean;
 }
 
-export interface IframeInitializerOptions extends InitializerOptions, IIframeInitializerOptions { }
+export interface IframeInitializerOptions extends InitializerOptions, IIframeInitializerOptions {}
 
 export async function iframeInitializer({
     communication,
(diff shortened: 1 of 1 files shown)


AviVahl:
Test? What does it fix?

AviVahl:
Btw using || instead of ?? feels wrong, that's why I'm asking.

packages/core/src/com/initializers/iframe.ts
```diff
@@ -40,19 +40,19 @@ export function deferredIframeInitializer({ communication: com, env: { env, endp
                 envReadyPromise,
                 instanceId,
                 src:
-                    src ?? managed
+                    src || managed
```
RomanYarik, line 43: maybe 
```suggestion
                    src ?? (managed ? ... )
```

RomanYarik (approved):

RomanYarik:
please add a description describing why this fix was needed

tomermes:
> Btw using || instead of ?? feels wrong, that's why I'm asking.

Committed Roman's suggestion to add "(" instead, and added a description to what it fixes.

AviVahl:
What about a test? This can regress in the future, if not covered.

RomanYarik:
@AviVahl it wasn't covered previously either, since we don't have tests which checks the src...

opening a PR for testing this

PR #909, opened 2021-08-23
refactor(scripts): use parent to fetch config in entrypoint

RomanYarik:
Currently each iframe environment fetches the config for itself.

We've seen that if an iframe wishes to load from a source which is not in the same base host as the parent, it cannot fetch the configurations properly.

Using parent window for 2 purpuses:
1. fetch the configuration for each new child environment
2. cache results, so we will avoid unneeded cache

Ideally, this should be solved using some middle service which manages that for all the browser environments (maybe a service worker), but it requires extra research


packages/scripts/src/create-entrypoint.ts
```diff
@@ -365,10 +364,29 @@ function loadConfigFile(filePath: string, scopedName: string, configEnvName: str
 //#endregion
 
 //#region configs
+function populateConfig(envName: string, staticBuild?: boolean, publicConfigsRoute?: string, config?: TopLevelConfig) {
+    return `${staticBuild ? importStaticConfigs() : ''}
+${staticBuild && config ? addOverrideConfig(config) : ''}
+
+${publicConfigsRoute ? getRemoteConfigs(publicConfigsRoute, envName) : ''}
+
+${publicConfigsRoute ? `${addConfigsEventListenerForParentEnvironments(publicConfigsRoute)}` : ''}`;
```
AdamYahid, line 373: Switch the order of registering the listener and posting the message

packages/scripts/src/create-entrypoint.ts
```diff
@@ -386,13 +404,47 @@ function importStaticConfigs() {
     }`;
 }
 
+function addConfigsEventListenerForParentEnvironments(publicConfigsRoute: string) {
+    return `if(isMainEntrypoint) {
+        const fetchedConfigs = {};
+        const configsEventListener = async ({ data: { id, envName }, source }) => {
+            if(source && id === '${publicConfigsRoute}') {
```
AviVahl, line 411: having so much untyped entry code in a string is problematic, and may cause maintenance burden.

the initial goal/approach for the entry generation code was to try and have it mostly typed. the generated entrypoint would ideally only pass build values to imported function calls. those functions (like `runEngineApp`) were themselves typed.

I suggest separating the parts that can be typed to ts files and importing their output in the entry code.

packages/scripts/src/create-entrypoint.ts
RomanYarik: it doesn't matter in this case. only the parent environment registers the listener, and only the child env will access the parent using this listener

packages/scripts/src/create-entrypoint.ts
RomanYarik: I do agree. We already have the same notion in the engine-runtime-node package.

I think it can be a good first issue @AdamYahid

tomermes (approved):

tomermes (approved):
Solved the issue with the full app (1 out of 3 "replaces")

packages/scripts/src/create-entrypoint.ts
AdamYahid: Technically your right, in this case  it doesn't because everything is async, but as a practice it's a potential race.

packages/scripts/src/create-entrypoint.ts
RomanYarik: created an issue https://github.com/wixplosives/engine/issues/911

PR #910, opened 2021-08-23
infra(core): add test for providing custom src

RomanYarik:


tomermes (approved):

packages/core/src/com/initializers/iframe.ts
```diff
@@ -33,11 +47,17 @@ export function deferredIframeInitializer({ communication: com, env: { env, endp
     const envReadyPromise = com.envReady(instanceId);
     return {
         id: instanceId,
-        initialize: ({ managed, iframeElement, hashParams, src }: IIframeInitializerOptions) => {
+        initialize: ({
+            managed,
+            iframeElement,
+            hashParams,
+            src,
+            envReady = envReadyPromise,
```
AviVahl, line 55: what's the benefit of getting this during the call just to return it in the return value?

AviVahl: just for the test?

RomanYarik: yes

AviVahl: how can it be avoided then? why is the environment not initializing during the test?
if we keep this, it's a probable source for false positives.

RomanYarik: because in the test I don't really load an engine environment, but a blob url with very small html inside.
What I can do, is to mock the ready message from that preview.

ill give it a try

PR #917, opened 2021-08-31
feat!: allow environment inheritance

nadav-ab:
allow environments to inherit from other 'multi' endpoint environments


packages/core/test/node/env-dependencies.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+import chaiAsPromised from 'chai-as-promised';
+import sinonChai from 'sinon-chai';
+import { Environment, Feature, run as runEngine, Service } from '@wixc3/engine-core';
+
+chai.use(sinonChai);
+chai.use(chaiAsPromised);
+
+describe('ENV dependencies', () => {
+    it('simple env dependency', async () => {
+        const baseEnv = new Environment('baseEnv', 'node', 'multi');
```
RomanYarik, line 11: an error should be thrown (and displayed maybe type-wise) if extendingEnv is not multi, or if it's node node

packages/runtime-node/src/node-environment.ts
```diff
@@ -6,12 +6,13 @@ import {
...
     FeatureLoadersRegistry,
     IPreloadModule,
+    Environment,
 } from '@wixc3/engine-core';
 import { init, remapToUserLibrary, clear } from './extrenal-request-mapper';
 
 import type { IEnvironment, StartEnvironmentOptions, IStaticFeatureDefinition } from './types';
 
-export async function runNodeEnvironment({
+export async function runNodeEnvironment<ENV extends Environment>({
```
RomanYarik, line 15: maybe in this case it's better to pass the environment and not the name, and to infer the env type inside runNodeEnvironment?

RomanYarik (commented):
is there any difference when trying to extend contextual environments?

packages/core/test/node/env-dependencies.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+        const extendingEnv = new Environment('extendingEnv', 'node', 'multi', [baseEnv]);
+        const entryFeature = new Feature({
+            id: 'entry',
+            api: {
+                service: Service.withType<{ increment: (n: number) => number }>().defineEntity(baseEnv),
+                service2:
+                    Service.withType<{ multiplyThenIncrement: (n: number) => number }>().defineEntity(extendingEnv),
+            },
+        });
+        const extendingFeature = new Feature({
```
AdamYahid, line 53: What role does extending feature play in this test?

packages/runtime-node/test/node-environments-manager.unit.ts
```diff
@@ -15,18 +23,27 @@ chai.use(chaiAsPromised);
...
     dependencies: [comEntry.scopedName],
     envFilePaths: {
         server: require.resolve('@fixture/engine-node/dist/feature/x.server.env'),
     },
-    exportedEnvs: [{ name: 'server', type: 'node' }],
+    exportedEnvs: [
+        {
+            name: 'server',
+            type: 'node',
+            env: serverEnv,
```
AdamYahid, line 44: if we now enforce passing envs, and not strings, why don't we pass only the env instances? in this case there is a mismatch between the serverEnv.name and the value in the key "name", which might cause us issues down the line

AdamYahid (approved):

RomanYarik:
close due to #963

PR #993, opened 2021-11-01
feat(test-kit): add allowedErrors for withFeature

AviVahl:
naive mechanism to allow certain error messages to not fail the tests. others will still fail as expected.

this is *not* as flexible as the previous handler api, and that's actually on purpose. we'd like to lean towards the strict side as much as possible to avoid "real" unhandled logic errors (as opposed to ones our applications want to passthrough on purpose).

### DIFF
--- a/packages/test-kit/src/with-feature.ts
+++ b/packages/test-kit/src/with-feature.ts
@@ -66,6 +66,12 @@ export interface IFeatureExecutionOptions {
      * Creates a playwright trace file for the test
      */
     tracing?: boolean | Tracing;
+
+    /**
+     * Error messages that are allowed to stay unhandled without failing the tests.
+     * strings are tested for exact match.
+     */
+    allowedErrors?: Array<string | RegExp>;
 }
 
 export interface IWithFeatureOptions extends Omit<IFeatureExecutionOptions, 'tracing'>, playwright.LaunchOptions {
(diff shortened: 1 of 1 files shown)


AviVahl:
@alisey per our discussion

@eladdo92 we've noticed the `preventDefault` caused dev tools not not treat these errors as unhandled exceptions, which is much worst than double printing, and is actually unacceptable. this pr is similar to #991, but a bit more naive and strict about what it allows. I believe it is sufficient for our current needs, and I can live with the tradeoffs and the smaller API surface changes. That being said, if the need rises to control this behavior more surgically during the lifetime of the test, I'd be willing to re-consider a more flexible solution as presented in #991

packages/test-kit/src/with-feature.ts
```diff
@@ -272,7 +280,13 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
                 page.setDefaultNavigationTimeout(30_000);
                 page.setDefaultTimeout(10_000);
                 page.on('pageerror', (e) => {
-                    capturedErrors.push(e);
+                    if (
+                        !allowedErrors.some((allowed) =>
+                            allowed instanceof RegExp ? allowed.test(e.message) : e.message.includes(allowed)
```
AviVahl, line 285: Readability is more important than performance here

alisey: `e.message.includes(allowed)`, but "strings are tested for exact match" in the API comment?

AviVahl: My bad will verify equality. Nice catch.

alisey (approved):

PR #997, opened 2021-11-04
feat(com): register environment on host message

RomanYarik:
Currently, when the host receives a message from an environment, the communication does not explicitly registers the environment with the host to it's own communication, and the mechanism relies on a fallback mechanism.

Keeping the fallback mechanism for specific browser behavior, but also adding a new message handler which maps the sender of the message to the host the message was received through.
This will "win" the default communication behavior and locate the proper host when responding to the message.


packages/core/src/com/message-types.ts
```diff
@@ -26,7 +26,8 @@ export interface CallbackMessage extends BaseMessage {
 
 export interface ListenMessage extends BaseMessage {
     type: 'listen';
-    data: RemoteCallAddress & { handlerId: string };
+    data: RemoteCallAddress;
+    handlerId: string;
```
AviVahl, line 30: apply to unlisten as well

packages/core/src/com/communication.ts
```diff
@@ -167,6 +168,18 @@ export class Communication {
      */
     public registerMessageHandler(target: Target): void {
         target.addEventListener('message', this.handleEvent, true);
+        const onTargetMessage = ({ data }: { data: null | Message }) => {
```
AviVahl, line 171: merge listeners

barak007 (changes requested):
I want to go over it together one more time.

packages/core/src/com/communication.ts
```diff
@@ -166,14 +168,26 @@ export class Communication {
      * Add local handle event listener to Target.
      */
     public registerMessageHandler(target: Target): void {
-        target.addEventListener('message', this.handleEvent, true);
+        const onTargetMessage = ({ data }: { data: null | Message }) => {
+            if (isMessage(data)) {
+                if (!this.environments[data.from]) {
+                    this.registerEnv(data.from, (target as BaseHost).parent ?? target);
```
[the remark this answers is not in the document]
RomanYarik: yes, for all the messages coming from this target

barak007 (approved):

PR #1016, opened 2021-11-17
fix(com): handle environment disconnection

nadav-ab:


RomanYarik (approved):

packages/engineer/src/feature/dev-server.dev-server.env.ts
```diff
@@ -196,6 +198,11 @@ devServerFeature.setup(
 
             app.get('/feature-graph', (req, res) => {
                 const featureName = req.query['feature-name'] as string;
+                if (!featureName) {
+                    res.statusCode = 403;
```
AviVahl, line 202: why forbidden? use 404

PR #1058, opened 2021-12-13
refactor(scripts): export config location function and unify usage

RomanYarik:
currently same code is duplicated across many projects, and each time the implementation is "similar but not quite the same" as the function in the config middleware, which had the original behavior.

moved this method to `@wixc3/engine-runtime-node`  since the node-environment-manager uses pretty much the same (worse) implementation.

### DIFF
--- a/packages/runtime-node/src/node-environments-manager.ts
+++ b/packages/runtime-node/src/node-environments-manager.ts
@@ -30,6 +30,7 @@ import {
 } from './types';
 import { resolveEnvironments } from './environments';
 import { IPCHost, LOCAL_ENVIRONMENT_INITIALIZER_ENV_ID } from '@wixc3/engine-core-node';
+import { DefaultConfigFileExports, findAllConfigs } from './config-locators';
 
 export interface OverrideConfig {
     configName?: string;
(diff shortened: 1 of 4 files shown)


packages/runtime-node/src/node-environments-manager.ts
```diff
@@ -424,28 +431,24 @@ export class NodeEnvironmentsManager {
...
     }
 
-    private async getConfig(configName: string | undefined) {
+    private async getConfig(configName: string | undefined, envName: string, basePath: string) {
         const config: TopLevelConfig = [];
         const { configurations } = this.options;
         if (configurations && configName) {
-            const configDefinition = configurations.get(configName);
-            if (!configDefinition) {
+            const foundConfig = await findAllConfigs({
```
AviVahl, line 438: `getConfig` now re-searches for configs on every call. is that really the behavior we want?

RomanYarik: this is exactly what happened before

```
 for (const definition of configDefinition) {
                try {
                    if (Array.isArray(definition)) {
                        config.push(...definition);
                    } else {
                        config.push(...((await import(definition.filePath)) as { default: TopLevelConfig }).default);
                    }
                } catch (e) {
                    console.error(e);
                }
            }
```

AviVahl (approved):

PR #1059, opened 2021-12-13
feat!(scripts): electron entries fetch configs through ipc

RomanYarik:
Currently in electron, in electron-renderer target dev mode, we use the dev server to fetch configs, and it is not possible to create a live configuration.

If the electron-renderer will fetch from parent (ipcMain) the config, we would be able to control it for the applications running in electron.

### DIFF
--- a/packages/scripts/src/create-entrypoint.ts
+++ b/packages/scripts/src/create-entrypoint.ts
@@ -2,7 +2,6 @@ import type { SetMultiMap, TopLevelConfig } from '@wixc3/engine-core';
 import type { IConfigDefinition } from '@wixc3/engine-runtime-node';
 import { parse } from 'path';
 import { CONFIG_QUERY_PARAM, FEATURE_QUERY_PARAM } from './build-constants';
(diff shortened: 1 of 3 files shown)


packages/scripts/src/create-entrypoint.ts
```diff
@@ -540,6 +554,10 @@ function fetchFeaturesFromElectronProcess(externalFeaturesRoute: string) {
     return `await require('electron').ipcRenderer.invoke('${externalFeaturesRoute}')`;
 }
 
+function fetchConfigsForElectron(configsRoute: string, envName: string) {
+    return `await require('electron').ipcRenderer.invoke('${LOADED_FEATURE_MODULES_NAMESPACE}.${configsRoute}', { env: '${envName}', featureName, configName })`;
```
RomanYarik, line 558: this basically means that the electron entry (in electron-main target), will add a handler which will return the configuration, instead of the dev-server handling it.

RomanYarik:
dropping this for the moment, this is not required.

PR #1072, opened 2021-12-21
fix(core-node): handle errors on process.send  

RomanYarik:


packages/core-node/src/ipc-host.ts
```diff
@@ -1,6 +1,10 @@
 import { BaseHost, IDisposable, Message } from '@wixc3/engine-core';
 import type { ChildProcess } from 'child_process';
 
+export const isParentProcess = (process: NodeJS.Process | ChildProcess): process is NodeJS.Process => {
```
zviktorz, line 4: Let's consider using `process.ppid` to check if process has parent process ID
```js
const isParentProcess = () => !!process.ppid
```

zviktorz (changes requested):

PR #1080, opened 2021-12-27
fix(script): resolve external requests using `@file-services/resolve`

RomanYarik:
currently we use `require.resolve` when trying to understand if the request is to a feature file when crating the webpack config for the external feature. 
This will fails when browser fields are configured.

Using `@file-services/resolve` since this resolver handles browser fields

### DIFF
--- a/packages/scripts/package.json
+++ b/packages/scripts/package.json
@@ -12,6 +12,7 @@
   },
   "dependencies": {
     "@file-services/node": "^5.4.3",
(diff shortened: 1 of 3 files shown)


barak007 (approved):
LGTM

AviVahl:
I personally think it's the wrong approach to even look at what path was resolved.

RomanYarik:
@AviVahl I agree. Opened an issue https://github.com/wixplosives/engine/issues/new

PR #1122, opened 2022-01-20
feat(engineer): expose application output path in run options

liorer:


RomanYarik:
there is another idea we can consider, which doesn't require this. let's discuss on Monday

RomanYarik:
Went with a different solution, closing.

PR #1127, opened 2022-01-24
feature: runtime metadata feature

liorer:
## Description
The `RuntimeMetadata` feature exposes a config populated with the following properties: 
* `devport` - running dev server (started by engineer/electron) port (for dev mode).
* `applicationPath` - path to the build directory (for production mode).

This feature is intended to be consumed by engine applications in order to enable the following options:
* Redirecting a server running on the user's application to dev server which is started from engineer/electron (for dev mode).
* Serving `applicationPath` statically (for production mode).


RomanYarik:
please add a description for the PR

liorer:
@RomanYarik done

packages/engineer/test/dev-server.unit.ts
```diff
@@ -76,9 +78,15 @@ describe('engineer:dev-server', function () {
...
         engine: RuntimeEngine;
         runtimeFeature: typeof devServerFeature;
+        outputPath: string | undefined;
         config: { featureName: string | undefined; port: number };
     }> => {
-        const { dispose, engine, devServerFeature } = await startDevServer({
+        const {
+            dispose,
+            engine,
+            outputPath: devOutputPath,
```
RomanYarik, line 87: why do we need to return the outputPath from this method?

packages/engineer/test/dev-server.unit.ts
```diff
@@ -586,6 +600,22 @@ describe('engineer:dev-server', function () {
         expect(pid1).to.eq(pid2);
         expect(pid1).to.eq(process.pid);
     });
+
+    it('runs app with the correct runtime metadata', async () => {
```
RomanYarik, line 604: please also make a test in scripts, to check that the proper output path is added when running an already built application

test-fixtures/engine-runtime-metadata/src/feature/x.main.env.ts
```diff
@@ -0,0 +1,11 @@
+import { socketClientInitializer } from '@wixc3/engine-core';
+import sampleFeature, { client, server } from './x.feature';
+
+sampleFeature.setup(client, ({ runtimeMetadata, run }, { COM: { communication } }) => {
+    document.body.textContent = '';
```
RomanYarik, line 5: not sure it's required

packages/engineer/test/dev-server.unit.ts
```diff
@@ -586,6 +600,22 @@ describe('engineer:dev-server', function () {
...
+            config: { port },
+        } = await setup({
+            basePath: engineRuntimeMetadataFixturePath,
+            featureName: 'engine-runtime-metadata/x',
+        });
+
+        const page = await loadPage(`http://localhost:${port}/main.html`);
+        const text = await getBodyContent(page);
+
+        expect(text).to.contain(`"devport":${port}`);
```
RomanYarik, line 616: since the text is a stringified json, we can parse it ans make a test

RomanYarik (changes requested):

packages/engineer/test/dev-server.unit.ts
liorer: It's needed for the expected value, see line 617

packages/engineer/test/dev-server.unit.ts
liorer: Sounds good, done

test-fixtures/engine-runtime-metadata/src/feature/x.main.env.ts
liorer: That's correct, removing

packages/engineer/test/dev-server.unit.ts
liorer: Added test

AviVahl:
these output paths everywhere make no sense to me. in start mode, there's no output on the actual drive.
also, in the context of running a node environment, what does "outputPath" means? lets discuss offline.

PR #1130, opened 2022-01-26
fix: Avoid multiple event handlers of the same instance

barak007:
Probably breaking change

### DIFF
--- a/packages/core/test/node/remote-service-com.spec.ts
+++ b/packages/core/test/node/remote-service-com.spec.ts
@@ -1,10 +1,12 @@
 import chai, { expect } from 'chai';
+import chaiAsPromised from 'chai-as-promised';
 import sinonChai from 'sinon-chai';
 import { stub } from 'sinon';
 
 import { BaseHost, Communication, declareComEmitter } from '@wixc3/engine-core';
 
 chai.use(sinonChai);
+chai.use(chaiAsPromised);
 
 class EmitterService {
     listeners = new Set<(s: string) => void>();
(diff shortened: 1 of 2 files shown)


packages/core/test/node/remote-service-com.spec.ts
```diff
@@ -43,32 +43,68 @@ describe('com emitter service', () => {
...
+        main2.registerAPI(emitterServiceId, api);
+
+        const proxy = main.apiProxy<EmitterService>(Promise.resolve({ id: 'main2' }), emitterServiceId, {
+            ...declareComEmitter<EmitterService>('on', 'off', 'removeAll'),
+        });
+
+        const testListenerStub = stub();
+
+        await proxy.on(testListenerStub);
+        await expectAsyncErrorMessage(async () => {
```
RomanYarik, line 89: why not use chai-as-promised?
expect(proxy.on(testListenerStub)).to.eventually.rejectWith('..')

RomanYarik (approved):

packages/core/test/node/remote-service-com.spec.ts
barak007: I forgot we have this. changed

RomanYarik (approved):

barak007:
@RomanYarik after running the test in `wcs` It seems that this will not break it. So we can probably merge it as patch.
There is still the possibility of this breaking during work but I'm not sure how to test it. 
@Ari-Kishon can we make sure we play with `wcs` after this merge and try to see if there is any new errors? (if there are we want to fix them quickly)

Ari-Kishon:
@PeterShershov 
should we validate and merge this before or after the release?

barak007:
@Ari-Kishon 
I did a lot more testing and everything looks good. I think it's much safer to release. 
One thing. I did all my tests on the web version and not electron but I don't expect it to be deferent.

PR #1137, opened 2022-01-31
refactor(test-kit): cleaner node messaging

AviVahl:
uses Node's on function to listen for incoming messages
simplifies logic a bit


AviVahl:
Will reopen once rebased

PR #1146, opened 2022-02-08
feat: expose more metadata and add a noLogs cli command in engineer

liorer:
* Adds `--no-log` flag to cli. if `true`, will disable all console logs coming from dev server.
* Exposes `featureName`, `foundFeatures` and `isWorkspace` via RutimeMetadata feature.

### DIFF
--- a/packages/core/src/runtime-metadata.feature.ts
+++ b/packages/core/src/runtime-metadata.feature.ts
@@ -9,9 +9,9 @@ export interface FoundFeatures {
 export interface RuntimeMetadataConfig {
     devport?: number;
     applicationPath?: string;
-    isWorkspace?: boolean;
     featureName?: string;
     foundFeatures?: FoundFeatures[];
+    isWorkspace?: boolean;
 }
 
 export default new Feature({
--- a/packages/core/src/runtime-metadata.feature.ts
+++ b/packages/core/src/runtime-metadata.feature.ts
@@ -1,9 +1,14 @@
+import type { INpmPackage } from '@wixc3/resolve-directory-context';
+import type { FeatureEnvDefinition } from '@wixc3/engine-scripts';
 import { Config } from './entities/config';
 import { Feature } from './entities/feature';
 
 export interface RuntimeMetadataConfig {
     devport?: number;
-    applicationPath: string;
+    applicationPath?: string;
+    packages?: INpmPackage[];
+    featureName?: string;
+    featureEnvDefinitions?: Record<string, FeatureEnvDefinition>;
 }
 
 export default new Feature({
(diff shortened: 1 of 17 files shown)


packages/core/src/runtime-metadata.feature.ts
```diff
@@ -1,9 +1,17 @@
...
+export interface FoundFeatures {
+    featureName: string;
+    configurations: string[];
+}
+
 export interface RuntimeMetadataConfig {
     devport?: number;
-    applicationPath: string;
+    applicationPath?: string;
+    isWorkspace?: boolean;
```
AviVahl, line 12: why is it even relevant whether the source-code of the app sat in a workspace or not? please make sure to avoid mixing build-time concerns with runtime ones.

packages/engineer/src/feature/dev-server.dev-server.env.ts
```diff
@@ -298,19 +304,24 @@ devServerFeature.setup(
 
             await Promise.all(compilationPromises);
 
-            const mainUrl = `http://localhost:${actualPort}/`;
-            if (featureName) {
-                console.log('Main application URL:', `${mainUrl}main.html`);
-            }
+            if (!noLogs) {
```
AviVahl, line 307: in which flow/mode would we want the *dev server* to not log?

packages/core/src/runtime-metadata.feature.ts
```diff
@@ -1,9 +1,17 @@
...
+    configurations: string[];
+}
+
 export interface RuntimeMetadataConfig {
     devport?: number;
-    applicationPath: string;
+    applicationPath?: string;
+    featureName?: string;
+    foundFeatures?: FoundFeatures[];
+    isWorkspace?: boolean;
```
RomanYarik, line 14: the only thing I think here is maybe add this config to a key called "engineerMetadata" or something like that? to be clear it's dev mode only data

RomanYarik (changes requested):

packages/engineer/src/feature/gui.dev-server.env.ts
```diff
@@ -49,11 +48,6 @@ guiFeature.setup(
                 outputPath: application.outputPath,
             })
         );
-
```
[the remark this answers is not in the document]
liorer: it was moved to `dev-server.dev-server.env.ts`, thought it'd make more sense to have all the logs in one place. Also that way we can disable those logs more easily. please lemme know if there are any implications to that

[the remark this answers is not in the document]
RomanYarik: if this passes as a config, this feature can also depend on it and receive the metadata config. this is the feature which adds the dashboard entry to the dev server, so it doesn't make sense to have the log somewhere else

packages/core/src/runtime-metadata.feature.ts
liorer: Renamed config to `engineerMetadataConfig`

RomanYarik (approved):

PR #1149, opened 2022-02-09
feat: overridable config loader

zviktorz:


packages/config-loader/src/config-loader.ts
```diff
@@ -0,0 +1,12 @@
+declare const __webpack_public_path__: string; 
+
+export async function loadConfig(
```
RomanYarik, line 3: why should this be in a separate package?

packages/config-loader/src/config-loader.ts
```diff
@@ -0,0 +1,12 @@
...
+    fileName: string, 
+    envName: string, 
+) {
+    const configFileName = envName ? `${fileName}.${envName}` : fileName;
+    const url = `${__webpack_public_path__}configs/${configFileName}.json`;
```
RomanYarik, line 8: the `configs` key should be passed to this method, since it's dynamic at build time

packages/scripts/src/create-entrypoint.ts
```diff
@@ -109,6 +110,7 @@ export function createMainEntrypoint({
     externalFeaturesRoute,
     eagerEntrypoint,
     featuresBundleName,
+    configLoader = '@wixc3/config-loader'
```
RomanYarik, line 113: if this is not in a separate package, you could just pass `join(__dirname, 'default-config-loader')`. The logic will work the same, no?

RomanYarik: also I think about the naming. configLoader implies to me a method or something.
Maybe we should consider something like `configLoaderModulePath`?

RomanYarik (changes requested):
Please add tests

packages/config-loader/src/config-loader.ts
zviktorz: can we discuss it?

packages/engineer/src/cli-commands.ts
```diff
@@ -143,6 +143,7 @@ export function buildCommand(program: Command) {
         .option('--publicConfigsRoute <publicConfigsRoute>', 'public route for configurations')
         .option('--external [true|false]', 'build feature as external', parseBoolean, false)
         .option('--eagerEntrypoints [true|false]', 'build feature as external', parseBoolean, false)
+        .option('--configLoaderModuleName [configLoaderModuleName]', 'custom config loader module name')
```
RomanYarik, line 146: maybe we should state here that this will only apply for static builds

RomanYarik (changes requested):
Tests are missing.
I have a suspicion it will not work if providing some other loader when it is provided as a request to some dev 3rd party, but I am not sure

packages/scripts/src/application.ts
```diff
@@ -793,6 +798,7 @@ export class Application {
             createWebpackConfig: isExternal ? createWebpackConfigForExternalFeature : createWebpackConfig,
             externalFeaturesRoute,
             eagerEntrypoint,
+            configLoaderModuleName,
```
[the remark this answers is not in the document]
zviktorz: We with Ari agreed that picking this param from CLI is enough for now

packages/scripts/test/application.unit.ts
```diff
@@ -613,6 +613,51 @@ describe('Application', function () {
...
+            const publicConfigsRoute = 'config';
+            const app = new Application({ basePath: staticBaseWebApplicationFixturePath });
+            await app.build({
+                featureName: 'static-base-web-application/base-web-application',
+                singleFeature: true,
+                configName: 'static-base-web-application/base',
+                publicConfigsRoute: '/config',
+                staticBuild: true,
+            });
+            // disposables.add(() => app.clean());
```
RomanYarik, line 627: why this is commented?

packages/scripts/test/application.unit.ts
```diff
@@ -613,6 +613,51 @@ describe('Application', function () {
...
+        });
+
+        it('overrides config', async () => {
+            const app = new Application({ basePath: staticBaseWebApplicationFixturePath });
+            await app.build({
+                featureName: 'static-base-web-application/base-web-application',
+                singleFeature: true,
+                configName: 'static-base-web-application/base',
+                staticBuild: true,
+                configLoaderModuleName: '@fixture/static-base-web-application-feature/dist/override-config-loader'
```
RomanYarik, line 648: can you replace it with a request to a file in a different package? which this feature doesn't use at runtime?

RomanYarik (changes requested):
Nice

RomanYarik (approved):

PR #1164, opened 2022-02-21
fix(test-kit)!: don't default to waitUntil: 'networkidle'

AviVahl:
users of `withFeature -> getLoadedFeature` can specify whatever `navigationOptions` they want for either per-suite or per-test level

### DIFF
--- a/examples/file-server/src/test/file-server.ix.ts
+++ b/examples/file-server/src/test/file-server.ix.ts
@@ -1,5 +1,6 @@
 import path from 'path';
 import { withFeature } from '@wixc3/engine-test-kit';
+import { waitFor } from 'promise-assist';
 import { expect } from 'chai';
 import { FileServerDriver } from './file-server-driver';
 import fileServerFeature, { SERVER_MARK, MAIN_MARK } from '../feature/file-server.feature';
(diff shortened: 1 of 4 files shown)


examples/file-server/src/test/file-server.ix.ts
```diff
@@ -41,11 +42,14 @@ describe('File Server Feature', function () {
...
         const { getMetrics } = await getLoadedFeature();
-        const metrics = await getMetrics();
-        expect(metrics.marks.length).to.eq(2);
-        const browserMark = metrics.marks.find((m) => m.name === MAIN_MARK);
-        const serverMark = metrics.marks.find((m) => m.name === SERVER_MARK);
-        expect(browserMark, 'should receive browser marks').not.be.undefined;
-        expect(serverMark, 'should receive server marks').not.be.undefined;
+
+        await waitFor(async () => {
+            const metrics = await getMetrics();
```
RomanYarik, line 47: how many of these changes repos down the line would have?

AviVahl: Depending on the assumptions used in tests.
With the latest revision, moving `navigationOptions` to the options object passed to `withFeature/getLoadedFeature`, this test would require an even smaller change: `navigationOptions: { waitUntil: 'networkidle' }` to be passed in. That would reproduce the previous behavior.

The thing is... network being idle doesn't _necessarily_ say all code executed and finished its work (or that metrics are ready). I left the `waitFor` in, as I feel it would make the test a bit more robust and harder to break.

PR #1172, opened 2022-02-22
fix(test-kit)!: don't default to waitUntil: 'networkidle'

AviVahl:
users of `withFeature -> getLoadedFeature` can specify whatever `navigationOptions` they want for either per-suite or per-test level


AviVahl:
awaiting a minor release, then this will merge as a major release.

PR #1191, opened 2022-03-06
chore: engine communication cleanups

barak007:
This is the start of the effort to move the communication to a separated package also this PR will enable features to access the engine in the setup function

* Pass the `RuntimeEngine` to the `SettingUpFeature`
* Removed `getCOM` from engine runtime (remove coupling prepare for communication split)
* Removed `runningEnvironmentName` (can be accessed via runtime engine)
* Introduce `Value` entity for local provided values

This is a major version

### DIFF
--- a/packages/core/src/entities/value.ts
+++ b/packages/core/src/entities/value.ts
@@ -1,0 +1,31 @@
+import type { RuntimeEngine } from '../runtime-engine';
+import { CREATE_RUNTIME, REGISTER_VALUE } from '../symbols';
+import type { EnvVisibility } from '../types';
+import { FeatureOutput } from './output';
+
+export class Value<T, ProvidedFrom extends EnvVisibility> extends FeatureOutput<
+    T,
+    T,
+    ProvidedFrom,
+    ProvidedFrom,
+    false
+> {
+    public static withType<T>() {
+        return {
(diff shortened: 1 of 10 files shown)


packages/core/src/entities/value.ts
```diff
@@ -0,0 +1,31 @@
+import type { RuntimeEngine } from '../runtime-engine';
+import { CREATE_RUNTIME, REGISTER_VALUE } from '../symbols';
+import type { EnvVisibility } from '../types';
+import { FeatureOutput } from './output';
+
+export class Value<T, ProvidedFrom extends EnvVisibility> extends FeatureOutput<
```
RomanYarik, line 6: I think it requires some explanation to why, and when, we want to use thios entity.
If I understand every service which in not .allowRemoteAccess() is Value?

barak007: It's kind of true but not a requirement. I wanted that Service will inherit form this but it creates more complex implementation. The Service class should be part of the communication when we extract it. Value is more basic.

barak007: I don't think we need to document it right since we don't have docs for the other entities.

I want to document everything soon.

RomanYarik (approved):

PR #1273, opened 2022-04-24
chore: update deps

barak007:
Also moved transport to `websocket`


packages/core/src/com/hosts/ws-client-host.ts
```diff
@@ -14,13 +14,13 @@ export class WsClientHost extends BaseHost {
         const { promise, resolve } = deferred();
         this.connected = promise;
 
-        this.socketClient = io(url, options);
+        this.socketClient = io(url, {...options, transports: ['websocket'], forceNew: true});
```
barak007, line 17: should it be behind?

packages/runtime-node/src/launch-http-server.ts
```diff
@@ -47,7 +47,7 @@ export async function launchEngineHttpServer({
         openSockets.add(socket);
         socket.once('close', () => openSockets.delete(socket));
     });
-    const socketServer = new io.Server(httpServer, { cors: {}, ...socketServerOptions });
+    const socketServer = new io.Server(httpServer, { cors: {}, ...socketServerOptions, transports: ['websocket'] });
```
barak007, line 50: should it be behind?

AviVahl (approved):

PR #1303, opened 2022-05-17
Make Socket.IO transport overridable

zviktorz:
This way we will have socket.io transport defaulted to websockets only with ability to override during application start up like following:
```js
app.run({
    ...options,
    runtimeOptions,
    socketServerOptions: {
        pingTimeout: 60000 * 10,
        transports: ['websocket', 'polling']
    }
})
```

### DIFF
--- a/packages/core/src/com/hosts/ws-client-host.ts
+++ b/packages/core/src/com/hosts/ws-client-host.ts
@@ -14,7 +14,7 @@ export class WsClientHost extends BaseHost {
         const { promise, resolve } = deferred();
         this.connected = promise;
 
-        this.socketClient = io(url, { ...options, transports: ['websocket'], forceNew: true });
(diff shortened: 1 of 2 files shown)


thehulke:
@zviktorz Looking good 
can you please add to the description the pattern to override the `transports`?

barak007:
why do we need deferent transport?
I think we have one more place with this configuration

zviktorz:
> why do we need deferent transport? I think we have one more place with this configuration

we have components studio installed as npm dependency crashing due to inability to connection to socket.io via polling. either way it is better to have customazability rather then not having it

barak007:
I don't mind the ability to config but don't mix the issue of the npm package with this (don't change the behavior for the npm package).
There is something else going on there...there is no reason not to use websocket transport.

barak007 (commented):
did you checked that no one is passing these options currently?
I don't want to change the default transport

PR #1313, opened 2022-05-22
docs(feature): start of documentation

thehulke:
WIP of the `Feature` documentation in effort to provide devs a better coding experience.

Some props are lacking documentation at this point and docs will be added later on.

### DIFF
--- a/packages/core/src/entities/feature.ts
+++ b/packages/core/src/entities/feature.ts
@@ -75,22 +75,66 @@ export class RuntimeFeature<T extends Feature, ENV extends AnyEnvironment> {
     }
 }
 
+/**
+ * Feature is a class that can be used to create a runtime feature. It is used to define the feature and its dependencies.
+ *
+ * @class Feature
+ * @template ID - Way to identify the feature. Provide a unique string or symbol.
+ * @template Deps - List of dependencies.
+ * @template API - API that the feature exposes.
+ * @template EnvironmentContext - Context that the feature is running in.
+ */
 export class Feature<
     ID extends string = string,
     Deps extends Feature[] = any[],
     API extends EntityRecord = any,
     EnvironmentContext extends Record<string, DisposableContext<any>> = any
 > {
+    /**
+     * Identifier of the feature.
+     *
+     * @type {Feature<ID, Feature[], API, EnvironmentContext>}
+     * @memberof Feature
+     */
     public asEntity: Feature<ID, Feature[], API, EnvironmentContext> = this;
+    /**
+     * Unique string or symbol that identifies the feature.
+     *
+     * @type {ID}
+     * @memberof Feature
+     */
     public id: ID;
+    /**
+     * Dependencies of the feature. This is an array of features.
+     *
+     * @type {Deps}
+     * @memberof Feature
+     */
     public dependencies: Deps;
+    /**
+     * What the feature exposes. This is an object with the key being the name of the api and the value being the api.
+     *
+     * @type {API}
+     * @memberof Feature
+     */
     public api: API;
+    /**
+     * Context that the feature is running in. Bound to the environment.
+     *
+     * @type {EnvironmentContext}
+     * @memberof Feature
+     */
     public context: EnvironmentContext;
 
     private environmentIml = new Set<string>();
     private setupHandlers = new SetMultiMap<string, SetupHandler<any, any, Deps, API, EnvironmentContext>>();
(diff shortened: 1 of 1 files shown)


packages/core/src/entities/feature.ts
```diff
@@ -75,30 +75,79 @@ export class RuntimeFeature<T extends Feature, ENV extends AnyEnvironment> {
     }
 }
 
+/**
+ * Feature is a class that can be used to create a runtime feature. It is used to define the feature and its dependencies.
+ *
+ * @class Feature
```
AviVahl, line 81: not needed when actual classes are used.

packages/core/src/entities/feature.ts
```diff
@@ -75,30 +75,79 @@ export class RuntimeFeature<T extends Feature, ENV extends AnyEnvironment> {
...
 export class Feature<
     ID extends string = string,
     Deps extends Feature[] = any[],
     API extends EntityRecord = any,
     EnvironmentContext extends Record<string, DisposableContext<any>> = any
 > {
+    /**
+     * Identifier of the feature.
+     *
+     * @type {Feature<ID, Feature[], API, EnvironmentContext>}
```
AviVahl, line 96: lets avoid this repetition. we already write the type in the code.
we should probably use tsdoc: https://tsdoc.org/

packages/core/src/entities/feature.ts
```diff
@@ -75,30 +75,79 @@ export class RuntimeFeature<T extends Feature, ENV extends AnyEnvironment> {
...
+ * @template EnvironmentContext - Context that the feature is running in.
+ */
 export class Feature<
     ID extends string = string,
     Deps extends Feature[] = any[],
     API extends EntityRecord = any,
     EnvironmentContext extends Record<string, DisposableContext<any>> = any
 > {
+    /**
+     * Identifier of the feature.
```
AviVahl, line 94: The purpose of this field is to point to `this`, but with a _type_ that has the `Deps` set back to generic `Feature[]`.
It's used to workaround typescript inlining the deep types of deps (and their deps, deeply) when generating `xyz.feature.d.ts`.

There should be an `@example showing how this is used.`

packages/core/src/entities/feature.ts
```diff
@@ -75,30 +75,79 @@ export class RuntimeFeature<T extends Feature, ENV extends AnyEnvironment> {
     }
 }
 
+/**
+ * Feature is a class that can be used to create a runtime feature. It is used to define the feature and its dependencies.
+ *
+ * @class Feature
+ * @template ID - Way to identify the feature. Provide a unique string or symbol.
```
AviVahl, line 82: it extends string, so no symbol.

packages/core/src/entities/feature.ts
```diff
@@ -75,30 +75,79 @@ export class RuntimeFeature<T extends Feature, ENV extends AnyEnvironment> {
...
+ * @template Deps - List of dependencies.
```
AviVahl, line 83: features this feature depends on (consumes their APIs)

packages/core/src/entities/feature.ts
```diff
@@ -75,30 +75,79 @@ export class RuntimeFeature<T extends Feature, ENV extends AnyEnvironment> {
...
     ID extends string = string,
     Deps extends Feature[] = any[],
     API extends EntityRecord = any,
     EnvironmentContext extends Record<string, DisposableContext<any>> = any
 > {
+    /**
+     * Identifier of the feature.
+     *
+     * @type {Feature<ID, Feature[], API, EnvironmentContext>}
+     * @memberof Feature
```
AviVahl, line 97: not needed.

AviVahl (changes requested):
lets use tsdocs instead of jsdocs. meaning, we shouldn't write a type more than once.
the most important thing, IMO, is a description that clarifies what the name of the entity couldn't.

packages/core/src/entities/feature.ts
```diff
@@ -75,30 +75,79 @@ export class RuntimeFeature<T extends Feature, ENV extends AnyEnvironment> {
...
         this.api = def.api || ({} as API);
         this.context = def.context || ({} as EnvironmentContext);
         this.identifyApis();
     }
-
+    /**
+     * Register a setup handler for the feature on a specific environment.
+     * @param env - Environment to register the handler on.
+     * @param setupHandler - Function that will be called when the feature is setup.
+     * @returns (optional) The api that the feature exposes.
```
AviVahl, line 149: it's not optional. if the feature has an api, it must by returned y setup

PR #1356, opened 2022-07-06
feat(extend-envs-api): Added new methods to work with registered environments

olehrakwix:
- Added ability to get registered environments using communication API
- Removed `disposedEnvironments` property in order to be able to relaunch disposed env

### DIFF
--- a/packages/core/src/com/communication.ts
+++ b/packages/core/src/com/communication.ts
@@ -74,7 +74,6 @@ export class Communication {
     private messageHandlers = new WeakMap<Target, (options: { data: null | Message }) => void>();
     private disposListeners = new Set<(envId: string) => void>();
     private callbackToEnvMapping = new Map<string, string>();
-    private disposedEnvironments = new Set<string>();
 
     constructor(
         host: Target,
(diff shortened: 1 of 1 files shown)


packages/core/src/com/communication.ts
```diff
@@ -339,6 +339,14 @@ export class Communication {
         return this.environments[envName]?.host;
     }
 
+    public getRegisteredEnvironments(): string[] {
```
cylon3035, line 342:
```suggestion
    public getRegisteredEnvironmentNames(): string[] {
```

packages/core/src/com/communication.ts
```diff
@@ -339,6 +339,14 @@ export class Communication {
         return this.environments[envName]?.host;
     }
 
+    public getRegisteredEnvironmentNames(): string[] {
+        return Object.keys(this.environments).filter((id) => id !== '*');
```
AviVahl, line 343: that's not the names of the environments. it's the instance id. what's the use case to need access to this list?

olehrakwix: ok, fixed the name of the method.
it is supposed to be used to get all running envs in order to clear them.

here is what I'm going to use this:
```
        const environmentsToDispose = communication.getRegisteredEnvironments();
        environmentsToDispose.forEach((environment) => {
            if (environment !== electronEnv.env) {
                communication.clearEnvironment(environment);
            }
        });
        communication.resetDisposedEnvironments();
```

packages/core/src/com/communication.ts
```diff
@@ -339,6 +339,14 @@ export class Communication {
...
+    }
+
+    public resetDisposedEnvironments(): void {
+        this.disposedEnvironments.clear();
```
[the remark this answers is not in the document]
olehrakwix: I agree, but that was added a long time ago in order to avoid subscribing to disposed environments events.

This is supposed to be used in order to reinitialize the application (e.g. like in VS code when you are opening another folder). This can reset all envs except the root one and reinitialize everything from a scratch.

olehrakwix:
@AviVahl , ok, I'll add some jsdocs and tests for this

AviVahl (approved):

PR #1362, opened 2022-07-18
fix(dashboard): dependency graph plus refactor

thehulke:
Bugfixes: 
* Dependency graph not shown on first load
* Dependency graph is now being evaluating based on URL params
* Dependency graph is stacking graphs if the dependency input not cleared entirely (empty string)

Refactor:
* Decoupled `Dashboard` components a bit, introducing two new ones:
    - `dependency-graph`
    - `sidebar`

The refactor purpose is in favor of improving the UX of the dashboard, having it easier to read

### DIFF
--- a/packages/dashboard/src/components/dashboard.tsx
+++ b/packages/dashboard/src/components/dashboard.tsx
@@ -29,7 +29,7 @@ export interface SelectedFeature {
     runtimeArguments?: string;
 }
 
-interface IParams {
+interface IDashboardParams {
     user_feature: string | undefined;
     user_config: string | undefined;
 }
@@ -38,7 +38,7 @@ export interface IDashboardContext {
     serverState: {
         featuresWithRunningNodeEnvs: RunningEngineFeature[];
     };
-    params: IParams;
+    params: IDashboardParams;
     setParams: (t: URLParamsValue<'user_feature' | 'user_config'>) => void;
 }
 
--- a/packages/dashboard/src/components/dashboard.tsx
+++ b/packages/dashboard/src/components/dashboard.tsx
@@ -131,9 +131,13 @@ export const Dashboard = React.memo<IDashboardProps>(function Dashboard({
             ((!params.user_feature && !configName) || (configName && params.user_config === configName))
     );
 
-    const handleSelectedFeature = useCallback((featureName?: string) => {
-        setSelectedFeature(featureName ? featureName : '');
-    }, []);
+    const handleSelectedFeature = useCallback(
+        (featureName = '') => {
+            setSelectedFeature(featureName);
+            setParams({ user_feature: featureName, user_config: params.user_config });
+        },
+        [params.user_config, setParams]
+    );
     serverState.featuresWithRunningNodeEnvs;
 
     return (
--- a/packages/dashboard/src/components/dashboard.tsx
+++ b/packages/dashboard/src/components/dashboard.tsx
@@ -125,7 +125,7 @@ export const Dashboard = React.memo<IDashboardProps>(function Dashboard({
     );
 
     useEffect(() => {
-        if (params.user_feature) selectedFeatureConfig(params.user_feature).catch(console.log);
+        if (params.user_feature) selectedFeatureConfig(params.user_feature).catch(console.warn);
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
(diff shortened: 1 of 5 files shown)


packages/dashboard/src/components/dashboard.tsx
```diff
@@ -42,15 +67,15 @@ export const Dashboard = React.memo<IDashboardProps>(function Dashboard({
...
 
     const configNames = useMemo(
         () => serverState.features[params.user_feature || '']?.configurations ?? [],
         [params.user_feature, serverState.features]
     );
     const [firstConfigName] = configNames;
-    const [showGraph, setShowGraph] = useState(false);
 
-    const [selectedFeatureGraph, setSelectedFeatureGraph] = useState<GraphData | null>(null);
+    // const [selectedFeatureGraph, setSelectedFeatureGraph] = useState<GraphData | null>(null);
```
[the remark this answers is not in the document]
thehulke: removed, thanks

olehrakwix (approved):

packages/dashboard/src/components/dashboard.tsx
```diff
@@ -1,12 +1,13 @@
...
+import React, { createContext, useCallback, useEffect, useMemo, useState } from 'react';
 import type { GraphData } from '../graph-types';
-import { style, classes } from './dashboard.st.css';
-import { RuntimeOptionsContainer, IRuntimeOption } from './runtime-options-container';
+import { isServerResponseMessage, RunningEngineFeature, ServerState } from '../server-types';
 import { ActionsContainer } from './actions-container';
-import { FeatureGraph } from './feature-graph';
-import { useUrlParams } from './dashboard-hooks';
+import { URLParamsValue, useUrlParams } from './dashboard-hooks';
+import { classes } from './dashboard.st.css';
```
barak007, line 6: move st.css to the bottom

packages/dashboard/src/components/dashboard.tsx
```diff
@@ -1,12 +1,13 @@
...
+import DependencyGraph from './dependency-graph/dependency-graph';
+import { FeaturesSelection } from './feature-selection';
+import { IRuntimeOption, RuntimeOptionsContainer } from './runtime-options-container';
+import Sidebar from './sidebar/sidebar';
```
barak007, line 10: if possible please use named exports

packages/dashboard/src/components/dashboard.tsx
```diff
@@ -28,6 +29,28 @@ export interface SelectedFeature {
     runtimeArguments?: string;
 }
 
+interface IDashboardParams {
+    user_feature: string | undefined;
```
barak007, line 33: is there a reason for the snake_case? we usually just using camel case.

barak007: also, what is the user referring to here? maybe we can remove the prefix user_

packages/dashboard/src/components/dependency-graph/dependency-graph.tsx
```diff
@@ -0,0 +1,53 @@
...
+    fetchGraphData: (featureName: string) => Promise<GraphData>;
+}
+
+const DependencyGraph: FC<IDependencyGraphProps> = ({ fetchGraphData }) => {
+    const [showGraph, setShowGraph] = useState(false);
+    const [selectedFeatureGraph, setSelectedFeatureGraph] = useState<GraphData | null>(null);
+
+    const { params } = useContext(DashboardContext);
+
+    const applyFeatureGraphData = useCallback(
```
barak007, line 16: no need for this callback just do it inside the effect.

packages/dashboard/src/components/dashboard.tsx
thehulke: @barak007 It's the default of the import organizer

packages/dashboard/src/components/dashboard.tsx
thehulke: @barak007 let's put an action item to rethink the naming of it, 
it addresses the selected feature & config that is broken down in the dashboard (not the one that necessarily running as root feature by engineer)

packages/dashboard/src/components/dashboard.tsx
barak007: there is no such thing...."default import organizer".

packages/dashboard/src/components/dashboard.tsx
barak007: not sure why an action item is needed here,,, it's a simple thing that we could think about since it popped up in this PR.
If you want I will not insist to change this now...but I will ping you in a week to see what's going on with this.

PR #1364, opened 2022-07-18
Handle url with pathname creating WsClientHost

zviktorz:
Make WsClientHost be able to collect to websocket server hosted at some pathname for example:  wss://manage.wix.com/_api/wcs-gateway

### DIFF
--- a/packages/core/src/com/hosts/ws-client-host.ts
+++ b/packages/core/src/com/hosts/ws-client-host.ts
@@ -11,10 +11,13 @@ export class WsClientHost extends BaseHost {
     constructor(url: string, options?: Partial<SocketOptions>) {
         super();
 
+        const { protocol, host, pathname: path, searchParams } = new URL(url);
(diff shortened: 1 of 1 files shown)


packages/core/src/com/hosts/ws-client-host.ts
```diff
@@ -11,10 +11,13 @@ export class WsClientHost extends BaseHost {
     constructor(url: string, options?: Partial<SocketOptions>) {
         super();
 
+        const { protocol, host, pathname: path, searchParams } = new URL(url);
+        const query = Object.fromEntries(searchParams as unknown as Iterable<any>);
```
AviVahl, line 15: Object.fromEntries(searchParams.entries())

zviktorz: I tried but for some reason typescript doesn't know that entries are there

image

PR #1375, opened 2022-08-03
chore(external features): deleted all relevant code for external feature

thehulke:
Deleted all code relevant for external feature, it's basically dead code.
When we will get to implement it, we will overhaul `engine-core` so it will be able to run features mid-flight.

Everything that was deleted in this PR is irrelevant for now.


packages/scripts/src/application.ts
```diff
@@ -228,8 +187,6 @@ export class Application {
             singleFeature,
             // should build this feature in external mode
             isExternal: external,
```
cylon3035, line 189: this flag as well

barak007:
In the end we need to go all the way around...all features should be external. I can see a way to achieve this is by first delete everything first but I would want to see what's going to replace it. 
The minimal working parts are:
1. allow to load features in the entry point (even from url param) 
2. document how to build and maintain features complaint with an already built application (externals in general)

If we only remove the support for "external features" to integrate with other node environments. this become a much easier task. 
then we could re-think the node/processes integration

barak007 (commented):
In the end we need to go all the way around...all features should be external. I can see a way to achieve this is by first delete everything first but I would want to see what's going to replace it. 
The minimal working parts are:
1. allow to load features in the entry point (even from url param) 
2. document how to build and maintain features complaint with an already built application (externals in general)

If we only remove the support for "external features" to integrate with other node environments. this become a much easier task. 
then we could re-think the node/processes integration 


About the PR itself I would recommend to review it in IDE sine we are removing stuff other related things might not come up in the diff

AviVahl (approved):
merge/rebase master to this branch and land this.

packages/engineer/src/cli-commands.ts
```diff
@@ -142,8 +142,6 @@ export function buildCommand(program: Command) {
         .option('--outDir <outDir>', 'output directory for the built application', 'dist-app')
         .option('--webpackConfig <webpackConfig>', 'path to webpack config to build the application with')
         .option('--publicConfigsRoute <publicConfigsRoute>', 'public route for configurations')
-        .option('--external [true|false]', 'build feature as external', parseBoolean, false)
-        .option('--eagerEntrypoints [true|false]', 'build feature as external', parseBoolean, false)
```
AviVahl, line 146: this has nothing to do with external. it has to do with how the dynamic import is marked (for the entry point engine-scripts creates). see 'eager' @ https://webpack.js.org/api/module-methods/#magic-comments

could have been kept, but not crucial

packages/electron-commons/test/initializers/electron-node.spec.ts
```diff
@@ -20,17 +20,15 @@ const setupRunningEnv = async ({
...
         env: serverEnv,
-        getApplicationMetaData: () =>
-            Promise.resolve({
-                featureName: testFeature.id,
-                basePath: testProjectPath,
-                outputPath: fs.join(testProjectPath, 'dist-app'),
-                nodeEntryPath: fs.join(testProjectPath, 'entry.js'),
-                config: [testFeature.use({ errorType: { type: errorMode, handleUncaught } })],
-                features: Array.from(features.entries()),
-                env: serverEnv,
```
AviVahl: is env not needed anymore?

packages/electron-node/src/initializers/browser-window.ts
```diff
@@ -1,8 +1,8 @@
+import { once } from '@wixc3/common';
```
AviVahl, line 1: change unrelated to PR? could have been done in a separate one

packages/engineer/src/feature/dev-server.dev-server.env.ts
```diff
@@ -1,26 +1,23 @@
-import type io from 'socket.io';
-import devServerFeature, { devServerEnv } from './dev-server.feature';
-import { TargetApplication } from '../application-proxy-service';
-import express from 'express';
+import { createDisposables } from '@wixc3/create-disposables';
```
AviVahl, line 1: the imports reordering makes the review of the actual changes much harder.

packages/scripts/src/application/application.ts
```diff
@@ -76,15 +61,9 @@ export class Application {
...
         const opts = defaults(options, buildDefaults);
-        const { config: _config, path: configPath } = await this.getEngineConfig();
-        const config: EngineConfig = defaults(_config, { externalFeatureDefinitions: [] as IExternalDefinition[] });
+        const { config: _config } = await this.getEngineConfig();
+        const config = defaults(_config, buildDefaults);
 
-        if (opts.external && !opts.featureName) {
-            throw new Error('You must specify a feature name when building a feature in external mode');
-        }
-        if (config.require) {
```
AviVahl, line 85: this is important. engine.config.js -> require are hooks that we use to install `@stylable/node/r` and others

packages/scripts/src/application/application.ts
```diff
@@ -93,104 +72,24 @@ export class Application {
...
-
-        if (opts.includeExternalFeatures && config.externalFeatureDefinitions.length) {
-            externalFeatures.push(
-                ...getExternalFeaturesMetadata(config.externalFeatureDefinitions, externalFeaturesBasePath)
-            );
-        }
-
-        // creating external-features json either way
-        this.writeExtFeaturesJson(opts, externalFeatures);
+        const sourceRoot = opts.sourcesRoot ?? opts.featureDiscoveryRoot ?? '.';
```
AviVahl, line 78: rename opts into something meaningful

packages/scripts/src/application/defaults.ts
```diff
@@ -1,13 +1,8 @@
 import type { IBuildCommandOptions } from './types';
 
-export const DEFAULT_EXTERNAL_FEATURES_PATH = 'external-features.json';
-
 export const buildDefaults = {
     mode: 'production',
```
AviVahl, line 4: not sure keeping this file now improves the readability, considering most code was removed

AviVahl (approved):

packages/engineer/src/cli-commands.ts
thehulke: IMO It's best to cleanup at this case, and re-implement if needed

PR #1390, opened 2022-08-10
feat: support cross origin iframe env

cylon3035:
This pr enables ability to initialise iframe environment on cross origin.
What was done to enable cross origin

* removed direct access to iframe window during initialisation
* removed `managed: false` iframe initialisation mode https://github.com/wixplosives/engine/pull/1390#discussion_r944258555
* communication between host and iframe using COM during initialisation 
  cross iframe can communicate only via postMessage, in order to talk to iframe our script must be loaded first to register message handlers. I used simple light weight communication only during initialisation, but seems redundant to have another abstraction since initialisation flow ends with `runEngineApp` and communication feature is enabled. So we can use our own `COM` with `WindowInitializerService` to get required data/subscription from iframe in cross origin manner after `envReady` message from iframe.
* feature/config information shall be provided on iframe query params
  previously iframe was pulling feature/config information from parent main window url. introducing cross origin iframes makes it ambiguous to find correct "topWindow". This concern is removed by pulling feature/config info from own iframe url, since this data is any way copied by default in `packages/core/src/com/initializers/iframe.ts` `defaultSourceFactory`
* removed race condition with simultaneous initialisation
   instanceId was previously set to window.name by initialiser with direct access, async initializer + window.name set caused race condition between competing initializers to same iframe that had to be handled  https://github.com/wixplosives/engine/pull/207/files. Passing instanceId was moved to query param that can be done cross origin and latest url wins so no race condition
* cleanup of closed iframes on cross origin without direct access
   without direct access iframe should use postMessage to communicate, postMessage called in `onUnload` handler is ignored on cross origin iframes https://bugs.chromium.org/p/chromium/issues/detail?id=964950 `pagehide` event is more reliable and sends messages reliably


packages/core/src/com/initializers/iframe.ts
```diff
@@ -119,41 +114,41 @@ async function startManagedIframe({
...
+        const url = new URL(src, window.location.href);
+        const search = new URLSearchParams(url.search);
+        search.set(INSTANCE_ID_PARAM_NAME, instanceId);
+        url.search = search.toString();
+
+        iframe.src = url.href;
         await Promise.race([waitForCancel, waitForLoad(iframe)]);
 
-        if (iframe.contentWindow !== contentWindow || iframe.contentWindow.location.href !== href) {
+        // bridge to communicate cross origin windows during initialization
```
cylon3035, line 132: it is possible to use COM after envReady. Maybe It can be not very bad

packages/core/src/com/initializers/iframe.ts
```diff
@@ -119,44 +120,50 @@ async function startManagedIframe({
...
+        const url = new URL(src, window.location.href);
+        const search = new URLSearchParams(url.search);
 
-        if (iframe.contentWindow !== contentWindow || iframe.contentWindow.location.href !== href) {
+        com.registerEnv(instanceId, contentWindow);
+        // previously instance id was set to iframe window by host, but
+        // setting name to iframe window is not possible on cross origin and lead to race condition
+        // when multiple engine frames try to initialize in to the same iframe
+        // instead instance id is set via query param
+        // no race because latest url wins + no cross origin problems
```
alisey, line 137: I don't think it's relevant to explain what was happening before.

packages/core/test/iframe.ts
```diff
@@ -7,7 +7,10 @@ import {
     HashParamsRetriever,
 } from './test-api-service';
 
-const com = new Communication(window, self.name);
+const options = new URLSearchParams(window.location.search);
+const instanceId = options.get(INSTANCE_ID_PARAM_NAME)!;
+const com = new Communication(window, self.name || instanceId);
```
alisey, line 12: Why do we use `self.name` if instanceId is always provided? And why do we need to set the iframe name at all?

packages/core/src/com/initializers/iframe.ts
```diff
@@ -18,10 +18,19 @@ export interface IIframeInitializerOptions {
      */
     hashParams?: string;
     /**
-     * if true, allows control over ifrsme content via hash parameters also.
+     * if true, allows control over iframe content via hash parameters also.
```
alisey, line 21: I still don't understand the difference between managed and unmanaged.

What kind of control do hash params allow? And who's to stop me from changing the iframe hash params even if I don't specify `managed: true`?

alisey (approved):

packages/core/src/com/initializers/iframe.ts
cylon3035: valid point. Since we do not use `managed:false` outside of engine tests and examples and there no additional reasons to keep it - can I just remove it?

alisey: Sounds good.

packages/core/src/com/initializers/iframe.ts
```diff
@@ -47,57 +48,28 @@ export function deferredIframeInitializer({ communication: com, env: { env, endp
...
-            return managed
-                ? startManagedIframe({
-                      ...baseStartIframeParams,
-                      iframe: iframeElement,
-                  })
-                : startIframe({
-                      ...baseStartIframeParams,
-                      host: iframeElement,
-                  });
+            return startManagedIframe(startIframeParams);
```
thehulke, line 60: Since `managed` removed on this PR, consider proper naming

packages/core/src/com/initializers/iframe.ts
```diff
@@ -119,44 +91,47 @@ async function startManagedIframe({
...
 
     try {
-        const href = new URL(src, window.location.href).href;
         const contentWindow = iframe.contentWindow;
-        contentWindow.name = instanceId;
-        com.registerEnv(instanceId, contentWindow);
-        contentWindow.location.replace(href);
-        await Promise.race([waitForCancel, waitForLoad(iframe)]);
+        const url = new URL(src, window.location.href);
+        const search = new URLSearchParams(url.search);
```
thehulke, line 101:
```suggestion
        const searchParams = new URLSearchParams(url.search);
```

packages/core/src/com/initializers/iframe.ts
```diff
@@ -47,57 +48,28 @@ export function deferredIframeInitializer({ communication: com, env: { env, endp
...
-    }
-    com.registerEnv(instanceId, win);
-    await injectScript(win, instanceId, src);
-    await envReadyPromise;
-    return instanceId;
-}
-
 const cancellationTriggers = new WeakMap<HTMLIFrameElement, () => void>();
 
 async function startManagedIframe({
```
thehulke, line 75: consider renaming since `managed` was removed

thehulke (approved):
LGTM

thehulke (approved):

packages/core/src/com/initializers/iframe.ts
```diff
@@ -1,27 +1,28 @@
...
-     * if launching iframe in "managed" mode, it will parse the hash params on the other hand and listen to them
+     * it will parse the hash params on the other hand and listen to them
      */
     hashParams?: string;
     /**
-     * if true, allows control over ifrsme content via hash parameters also.
-     * @default false
+     * target host for iframe
+     * can be used to isolate environment on cross origin
+     * works only with managed: true
```
AviVahl, line 21: remove

packages/core/src/com/initializers/iframe.ts
```diff
@@ -119,44 +85,47 @@ async function startManagedIframe({
...
 
     try {
-        const href = new URL(src, window.location.href).href;
         const contentWindow = iframe.contentWindow;
-        contentWindow.name = instanceId;
-        com.registerEnv(instanceId, contentWindow);
-        contentWindow.location.replace(href);
-        await Promise.race([waitForCancel, waitForLoad(iframe)]);
+        const url = new URL(src, window.location.href);
+        const searchParams = new URLSearchParams(url.search);
```
AviVahl, line 95: url.searchParams

packages/core/test/delayed-iframe.ts
```diff
@@ -1,7 +1,10 @@
-import { Communication } from '@wixc3/engine-core';
+import { Communication, INSTANCE_ID_PARAM_NAME, WindowInitializerService } from '@wixc3/engine-core';
 import { TestService } from './test-api-service';
 
 setTimeout(() => {
-    const com = new Communication(window, self.name);
+    const options = new URLSearchParams(window.location.search);
```
AviVahl, line 5: window.location.searchParams.get

packages/core/test/iframe.ts
```diff
@@ -7,7 +7,15 @@ import {
     HashParamsRetriever,
 } from './test-api-service';
 
-const com = new Communication(window, self.name);
+const options = new URLSearchParams(window.location.search);
```
AviVahl, line 10: ditto

test-fixtures/cross-iframe/src/feature/configured-iframe.main.env.ts
```diff
@@ -0,0 +1,43 @@
+import { iframeInitializer } from '@wixc3/engine-core';
+import { mainEnv, iframeEnv } from './echo.feature';
+import fixture from './configured-iframe.feature';
+
+fixture.setup(mainEnv, ({ run, config }, { COM: { communication }, echoFeature: { echoService } }) => {
+    const myFrame = document.createElement('iframe');
+    myFrame.id = 'iframe';
+    document.body.append(myFrame);
+    const fConfigs = [
+        { host: '127.0.0.1:3000', param: 'abc' },
```
AviVahl, line 10: lets try to have an automated test that doesn't have that hard-wired

AviVahl (approved):

packages/core/src/com/initializers/iframe.ts
cylon3035: window.location has only `search` as `string`, no `searchParams`

packages/core/test/delayed-iframe.ts
cylon3035: window.location has only `search` as `string`, no `searchParams`

PR #1397, opened 2022-08-16
feat(node-environment-initializer): Updated node environment initializer to be synchronous

olehrakwix:
Changed function `initializeNodeEnvironment` to be synchronous instead of async.
The callers might need this to have immediate access to the `dispose` function.
This is needed cause, in some cases, application exit might be requested in between two moments: the environment is already spawned, and the environment did not send a callback via communication yet. Example use case: close electron application in one second after window appear. Because closing the window causes the host for the electron-renderer process to be disposed (at least I think so), the `envReady` event is not triggered via communication for the newly created environment.
In that case, the caller will never be able to dispose such env cause it will not be registered, and dispose function will not be returned. However, the process will be already spawned. In that case, the caller will have running background processes with those environments, which is not good.


AviVahl (approved):
The description is awesome. This is a breaking api change so major

PR #1412, opened 2022-08-28
feat: add flag that enabled/disables signing of the package

AlexShemeshWix:
Add flag that enables signing of the electron package


packages/electron-scripts/src/commands/build.ts
```diff
@@ -120,6 +125,15 @@ export async function build(options: IBuildCommandOptions): Promise<void> {
 
     const builderConfig = fs.readJsonFileSync(fs.join(basePath, electronBuilderConfigFileName)) as Configuration;
 
+    if (!sign) {
+        // Disable code signing on windows
+        if (builderConfig.win) {
+            Object.assign(builderConfig.win, {certificateSubjectName: null});
```
AviVahl, line 131: @AlexShemeshWix  Seeing this again, I think we should just allow .js config for builder. Changing process.env and assigning to a readonly field feel too hacky.

packages/electron-scripts/src/commands/build.ts
```diff
@@ -118,8 +118,9 @@ export async function build(options: IBuildCommandOptions): Promise<void> {
         outDir,
     });
 
-    const builderConfig = fs.readJsonFileSync(fs.join(basePath, electronBuilderConfigFileName)) as Configuration;
-
+    const configFullPath = fs.resolve(fs.join(basePath, electronBuilderConfigFileName));
+    const builderConfig: Configuration = await import(configFullPath) as Configuration;
```
AviVahl, line 122: when we dynamically import a commonjs file, we get it as a default export.
so `const {default: builderConfig} = (await import(configFilePath)) as {default: Configuration}`

AviVahl, line 122: when we dynamically import a commonjs file, we get it as a default export.
so `const {default: builderConfig} = (await import(configFilePath)) as {default: Configuration}`

AviVahl (changes requested):

AviVahl (approved):
please use prettier and ran it on the new code
the comma should be there. there's missing spaces and semicolon.
after doing that, feel free to merge to master

PR #1430, opened 2022-09-12
fix(events): unsubscribe handlers not being invoked

### LINKED ISSUE #1423: Com: unsubscribe handlers in services are never invoked
If you have a service:

```ts
class FooService {
    subscribe(handler: () => void) {}
    unsubscribe(handler: () => void) {}
}
```

Declared in the feature file as:
```ts
Service
    .withType<FooService>()
    .defineEntity(node)
    .allowRemoteAccess(declareComEmitter<FooService>('subscribe', 'unsubscribe'))
```

Then invoke it from a different environment as:
```ts
await FooService.subscribe(noop);
await sleep(1000);
await FooService.unsubscribe(noop);
```

Will not actually call the `unsubscribe` function.

The problem is this piece of code:
https://github.com/wixplosives/engine/blob/40023882a1b79d88b04879bc89d0309bd55188b1/packages/core-node/src/ws-node-host.ts#L65-L67

When adding / removing the handler from the list of listeners, a different handler id is used for `listen` and `unlisten`.

VladyslavG:
fixes #1423

The main problem was that we were namespacing `handlerId` in the `subscribe` calls and not in the `unsubscribe`.
Second problem was that we were doing it inside the Node Host - as a result it was not obvious inside the `Communication` whether `handlerId` is namespaced or not in each method.

So there are 2 fixes:
1) namespace `handlerId` in both `subscribe` and `unsubscribe`
2) add transparent namespacing into `Communication`

### DIFF
--- a/packages/core-node/src/ws-node-host.ts
+++ b/packages/core-node/src/ws-node-host.ts
@@ -29,11 +29,7 @@ export class WsServerHost extends BaseHost implements IDisposable {
         if (data.to !== '*') {
             if (this.socketToEnvId.has(data.to)) {
                 const { socket, clientID } = this.socketToEnvId.get(data.to)!;
-                const tenantId = data.to;
                 data.to = clientID;
-                if (data.type === 'event') {
-                    data.handlerId = data.handlerId.slice(0, data.handlerId.length - tenantId.length);
-                }
                 socket.emit('message', data);
             } else {
                 this.server.emit('message', data);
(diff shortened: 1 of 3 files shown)


alisey (approved):
Looks good to me. What about `forwardListen` and `forwardUnlisten`, don't they need the same treatment?

Please run prettier, there's a couple of missing semicolons.

VladyslavG:
> Looks good to me. What about `forwardListen` and `forwardUnlisten`, don't they need the same treatment?
> 
> Please run prettier, there's a couple of missing semicolons.

prettier- ok

`forwardListen` and `forwardUnlisten` currently work well. There were no attempts to namespace handlers for them.
However it could be explained by the fact that they were not used in scenarios where that would be needed. Potentially such scenario can occur in the future.

Do you think we need to namespace handlers for them @alisey ?

alisey:
@VladyslavG ¯\_(ツ)_/¯

If everything works well, let's merge as is.

barak007 (approved):
Looks awesome.

PR #1439, opened 2022-09-20
fix: changed dispose to be  parallel

daomry:
related to issue https://github.com/wixplosives/component-studio/issues/11447

### DIFF
--- a/packages/core/src/run-engine-app.ts
+++ b/packages/core/src/run-engine-app.ts
@@ -59,9 +59,7 @@ export function runEngineApp<ENV extends AnyEnvironment>({
         engine,
         async dispose() {
             await runningPromise;
-            for (const feature of features) {
-                await engine.dispose(feature);
-            }
+            await Promise.all(features.map(f => engine.dispose(f)));
         },
     };
 }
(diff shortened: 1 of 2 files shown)


daomry:
see also https://github.com/wixplosives/core3-utils/pull/123

daomry:
Note: this fix only gets the time down to 25 sec on my machine, the core issue is in closing  our webpack middleware. I'm still investigating that

YardenPorat:
I'm not sure if the disposal order is not important. @thehulke  can you assist?

thehulke:
@daomry is this behavior tested against WCS?

IMO order here is important, since you cannot control the flow it can lead to all sorts of issues.

YardenPorat:
I wonder who holds it up when it's sequential. Maybe we can console.time the dispose time on some debug logging

daomry:
the holdup is in the shutdown of the webpack middleware of engine's devserver, specifically closing watches. see
https://github.com/wixplosives/engine/blob/4b3bee54f222129a566db7599e484a2cb22b105b/packages/engineer/src/feature/dev-server.dev-server.env.ts#L241-L246
 it's likely due to open handles within the electron up but I'm still investigating

PR #1444, opened 2022-09-22
chore: refactor utils

daomry:


packages/engineer/package.json
```diff
@@ -24,6 +24,7 @@
     "@wixc3/engine-runtime-node": "^29.0.1",
     "@wixc3/engine-scripts": "^29.0.1",
     "commander": "^9.4.0",
+    "lodash": "^4.17.21",
```
AviVahl, line 27: we decided to not include `lodash` in our packages.
it is unmaintained, and when it had security vulnerabilities they ignored PRs for months.
a lot of its functionality can be achieved via other means (new native syntax, or other libraries)

daomry: good to know. is there a similar lib in use in our code?

AviVahl: not a single library that contains many utils.
we got away we native syntax for most things

_shallow_ defaults, for example, is as simple as `{...a, ...b}`, while `unique` is just `Array.from(new Set(items))`, etc.

daomry: great,  but I'm using defaultsDeep, which I need to implement and test... underscore.js?

daomry: removed lodash dependency, now depends on https://github.com/wixplosives/core3-utils/pull/138

PR #1452, opened 2022-09-28
fix(communication): Fixed circular message forwarding

olehrakwix:
Added `forwardingChain` property to the message object so we can detect and fail on circular message forwarding.

When a message is forwarded from one env to another, we add the current instance id to a `forwardingChain` array, so we can detect if the same message comes to our instance circularly and fail this message.

### DIFF
--- a/packages/core/src/com/communication.ts
+++ b/packages/core/src/com/communication.ts
@@ -175,7 +175,8 @@ export class Communication {
                                 method,
                                 args,
                                 this.rootEnvId,
-                                serviceComConfig as Record<string, AnyServiceMethodOptions>
+                                serviceComConfig as Record<string, AnyServiceMethodOptions>,
+                                []
                             );
                         // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
                         obj[method] = runtimeMethod;
@@ -233,7 +234,8 @@ export class Communication {
         method: string,
         args: unknown[],
         origin: string,
-        serviceComConfig: Record<string, AnyServiceMethodOptions>
+        serviceComConfig: Record<string, AnyServiceMethodOptions>,
+        forwardingChain: string[]
     ): Promise<unknown> {
         return new Promise<void>((res, rej) => {
             const callbackId = !serviceComConfig[method]?.emitOnly ? this.idsCounter.next('c') : undefined;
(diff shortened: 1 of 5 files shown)


packages/core/src/com/communication.ts
```diff
@@ -449,14 +458,28 @@ export class Communication {
...
+            origin: message.to,
+            type: 'callback',
+            forwardingChain: [this.rootEnvId],
+        };
+
+        if (message.forwardingChain.indexOf(this.rootEnvId) > -1) {
+            this.sendTo(message.from, {
+                ...responseMessage,
+                error: new Error(
+                    `cannot reach environment '${message.to}' from '${message.from}' as stuck in circular message forwarding`
```
thehulke, line 474:
```suggestion
                    `cannot reach environment '${message.to}' from '${message.from}' since it's stuck in circular messaging loop`
```

thehulke (approved):

packages/core/src/com/communication.ts
```diff
@@ -252,6 +258,7 @@ export class Communication {
                     data: { api, method, args },
                     callbackId,
                     origin,
+                    forwardingChain: [this.rootEnvId],
```
cylon3035, line 261: why call needs  `forwardingChain: [this.rootEnvId]` when it's direct call?

packages/core/src/com/communication.ts
```diff
@@ -587,6 +612,7 @@ export class Communication {
                     },
                     handlerId: listenerHandlerId,
                     origin,
+                    forwardingChain: [],
```
cylon3035, line 615: i think forwarding chain here should be populated

packages/core/src/com/communication.ts
```diff
@@ -618,6 +644,7 @@ export class Communication {
                         handlerId: this.createHandlerRecord(envId, api, method, fn),
                         callbackId,
                         origin,
+                        forwardingChain: [],
```
cylon3035, line 647: i think forwarding chain here should be populated

packages/core/src/com/communication.ts
cylon3035: after some thinking process i think `addOrRemoveListener`, `callMethod`, `forwardUnlisten` and `forwardListenmessage` should accept `forwardingChain` passed by caller

cylon3035: this way `forwardingChain` will be populated only inside `forwardMessage` and all functions mentioned above will respect and pass through `forwardingChain` value

olehrakwix: thanks, that really make sense. updated

cylon3035 (approved):

PR #1453, opened 2022-09-30
fix(deps): Reverted electron and playwright versions update

olehrakwix:
Revert major and minor version update for electron and playwright


cylon3035 (approved):

AviVahl:
Why are these two being reverted?

olehrakwix:
@AviVahl , because with new versions there are issues in WCS... electron tests are failing for unknown reasons, and there is no playwright installed on CI agent:
https://github.com/wixplosives/component-studio/actions/runs/3152189057/jobs/5127085763

AviVahl:
Electron version here is a devDep, and shouldn't affect which version is used by consumers. Missing playwright binary happens when the installed version of playwright-chromium does not match playwright-core. Mixing of versions can cause this.

AviVahl:
Also this repo is public. Avoid referring to private projects.

PR #1470, opened 2022-10-12
chore: refactor "analyze-feature" 

daomry:
Step 1: Prep
Updated imports to @wix/common
Broke out analyze-feature to a folder and split to different files


daomry:
*NOTE* https://github.com/wixplosives/core3-utils/pull/138 is a prerequisite

PR #1478, opened 2022-10-13
chore: cleanups

daomry:


examples/preload/src/test/utils.ts
```diff
@@ -7,15 +7,15 @@ export interface StartServerNewProcessOptions {
     featureName: string;
     runtimeOptions?: Record<string, string | boolean>;
     launchOptions?: playwright.LaunchOptions;
-    includeGui?:boolean;
+    minimal?:boolean;
```
thehulke, line 10: What does `minimal` stands for?

packages/engineer/src/feature/dev-server.dev-server.env.ts
```diff
@@ -282,6 +282,7 @@ devServerFeature.setup(
              *  but we will keep on watching on the users application
              *  2. the createCompiler function is not extendable with more configs with the current API
              */
+            console.time('Bundling files')
```
thehulke, line 285: Should the `console.log` exist? 
if so, consider being more specific about the task (IMO "files" is quite vague)

thehulke (commented):
Feeling a bit unsure regarding the usage of `minimal`
Can you please elaborate on what it stands for?

PR #1496, opened 2022-11-06
chore: updated dependencies @wixc3/common, /patterns and promise-assist

daomry:


package.json
```diff
@@ -19,6 +19,10 @@
     "test:unit": "lerna run test --stream --concurrency=1",
     "prettify": "npx prettier . --write"
   },
+  "dependencies": {
```
AviVahl, line 22: avoid adding deps in root. only dev-deps.
see sample-monorepo readme

packages/core/src/entities/feature.ts
```diff
@@ -263,7 +264,7 @@ export class Feature<
                     continue;
                 }
                 for (const key of Object.keys(featureOutput)) {
-                    settingUpFeature[key] = (featureOutput as any)[key];
+                    settingUpFeature[key] = (featureOutput)[key as keyof typeof featureOutput];
```
thehulke, line 267: Is casting necessary here?

daomry: since Object.keys(featureOutput) returns string[], the casting is required. 
I think casting the key has more intent than casting as any

AviVahl:
applied prettify to master. merge it to reduce diff size.

PR #1500, opened 2022-11-07
chore: ci updates

daomry:
- bumped CI node to 18
- dropped mac tests


AviVahl (changes requested):
we must work on Mac. we need CI to test it.
it only became flaky after the many recent refactors, and works well in many others repositories.

PR #1521, opened 2022-11-21
feat!: public path default set to empty

cylon3035:
public path is set to '' now by default.

Why?
* in codux I want to use bundled files from root and from path, thus I want relative assets

Why don't codux passes publicPath: '' instead of change in default param?
* there is no reason to have  / by default

### DIFF
--- a/packages/engineer/src/cli-commands.ts
+++ b/packages/engineer/src/cli-commands.ts
@@ -17,7 +17,7 @@ import type { ServerListeningHandler } from './feature/dev-server.types';
 
 const parseBoolean = (value: string) => value === 'true';
 const collectMultiple = (val: string, prev: string[]) => [...prev, val];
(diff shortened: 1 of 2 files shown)


AviVahl:
I think we should default to not passing anything when nothing was provided, rather than assume one or the other.

I think webpack@5 has special auto detection of public path when nothing is provided.

cylon3035:
> There's a special 'auto' value we might be able to use.

As I understand 'auto' option does not work with `__webpack_public_path__` in entry. Everything that this entry will load will have relative path `auto`.

Screenshot 2022-11-21 at 15 53 32

 If value is not provided, seems that webpack is using `/`. Documentation is not static clear what is default anyway.

AviVahl (approved):
Then an empty string is a better default for us. is this the only place that forces it?

PR #1541, opened 2022-12-04
feature: added consoleLogAllowedErrors to withFeature

daomry:


AviVahl:
I'm not sure I understand the intention of this PR. The name is confusing, as the error is an unhandled error or rejection on the page. Not something that was printed using `console.<log/error/warn>`. We later choose to print it using `console.error`, but that has nothing to do with what's happening within the page.

There's the `allowedErrors` field which allows us to ignore specific errors (for error checking scenarios). All other ones fail the test, like we wanted.

AviVahl:
Ah, I understand now.

AviVahl (approved):

PR #1562, opened 2022-12-17
feature: changed headless to be true if env.DEBUG=true

daomry:


packages/test-kit/src/with-feature.ts
```diff
@@ -165,6 +162,12 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
         navigationOptions: suiteNavigationOptions,
     } = withFeatureOptions;
 
+    const {
```
AviVahl, line 165: not sure why this got separated from the original destructuring..?

packages/test-kit/src/with-feature.ts
```diff
@@ -165,6 +162,12 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
         navigationOptions: suiteNavigationOptions,
     } = withFeatureOptions;
 
+    const {   
+        headless=!process.env.DEBUG,
```
AviVahl, line 166: if I only pass `devtools: true` from the `withFeature` call, I'd expect `headless` to be implcitly set to `false` (playwright's default behavior). Is us defining a default `true` overrides that behavior?

also, can we use something less generic than `DEBUG`? I'd expect that env variable to possibly affect other executables on some systems.

do we want to allow `DEBUG=false` or `DEBUG=0` to turn that off?

daomry: I was thinking the default is the current behavior , but no code changed needed for debugging.
changes to the code should override it.
re the env var name: TEST=true ? TEST_DBG=true?
I don't think DEBUG=false or DEBUG=0 should have an effect, only DEBUG=true. I'll adjust the test acordigly

thehulke (approved):

AviVahl (changes requested):
I just tested playwright's `chromium.launch` with `headless: true, devtools: true` and it doesn't work the same as `devtools: true`. The window wouldn't open.

daomry:
> I just tested playwright's `chromium.launch` with `headless: true, devtools: true` and it doesn't work the same as `devtools: true`. The window wouldn't open.

Why is it relevant here? if headless it true, devtools will be false in this situation.
The one nuance is that the default of devtools should probably be false unless headless: true (I'll update accordingly)

PR #1590, opened 2023-01-09
fix: messages to removed environments should be ignored

daomry:


barak007 (approved):
We could have dome it with null and if...but it really does not matter.

barak007 (approved):

PR #1594, opened 2023-01-10
fix(service): better error message

YardenPorat:
fix grammer, add a hint for possible issue

### DIFF
--- a/packages/core/src/entities/service.ts
+++ b/packages/core/src/entities/service.ts
@@ -63,7 +63,10 @@ export class Service<
             const localEnv = communication.getEnvironmentName();
             if (providedFrom.has(localEnv) || providedFrom.has(Universal.env)) {
                 if (!providedValue) {
-                    throw new Error('service is not provide in runtime');
+                    throw new Error(
+                        `Service is not provided in runtime.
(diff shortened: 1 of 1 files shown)


PeterShershov (approved):

packages/core/src/entities/service.ts
```diff
@@ -63,7 +63,10 @@ export class Service<
             const localEnv = communication.getEnvironmentName();
             if (providedFrom.has(localEnv) || providedFrom.has(Universal.env)) {
                 if (!providedValue) {
-                    throw new Error('service is not provide in runtime');
+                    throw new Error(
+                        `Service is not provided in runtime.
+                        Make sure environment setup file exist and correctly typed: [featureName].[envName].env.[ext]`
```
alisey, line 68: Speaking of grammar.

in runtime -> at runtime
sure environment -> sure the environment
exist -> exists
correctly typed -> named correctly

alisey: Can we also add the service name and the environment name to the error message? I can see we have `serviceKey` and `this.providedFrom` variables here.

YardenPorat: #1596

PR #1596, opened 2023-01-11
fix(service): add details to error msg

YardenPorat:


packages/core/src/entities/service.ts
```diff
@@ -64,8 +64,12 @@ export class Service<
             if (providedFrom.has(localEnv) || providedFrom.has(Universal.env)) {
                 if (!providedValue) {
                     throw new Error(
-                        `Service is not provided in runtime.
-                        Make sure environment setup file exist and correctly typed: [featureName].[envName].env.[ext]`
+                        `Service is not provided at runtime.
+Make sure the environment setup file exists and named correctly: [featureName].[envName].env.[ext]
+Service name: ${entityKey}
+Feature id: ${featureID}
+Environment: ${localEnv}
```
thehulke, line 71: @YardenPorat check identation please

YardenPorat: wdym check?

YardenPorat: A template string takes into account indentation
image

alisey (approved):

PR #1599, opened 2023-01-12
feat(with-feature): option for multiple browsers

matangeorgi:
An option to run multiple browsers within the withFeature, using env vars.
Chromium will run as the default browser, when no env vars are being passed.


packages/test-kit/src/with-feature.ts
```diff
@@ -149,6 +149,7 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
     const disposeAfterEach = createDisposables();
     const envDebugMode = 'DEBUG' in process.env;
     const debugMode = !!process.env.DEBUG;
+    const browserToRun = process.env.BROWSER as 'chromium' | 'firefox' | 'webkit' | undefined;
```
daomry, line 152: ts will not check it in runtime, check that the value ius a supported browser and throw a nice error message with a list of valid options

daomry (changes requested):

packages/test-kit/src/with-feature.ts
matangeorgi: Done :)

daomry (approved):
Well done. I moved the browser related code to its own file and made slight adjustments, let me know what you think

packages/test-kit/src/supported-browsers.ts
```diff
@@ -0,0 +1,13 @@
...
+export type SupportedBrowser = typeof supportedBrowsers[number];
+
+export function isValidBrowserName(browserName: string): browserName is SupportedBrowser {
+    return (supportedBrowsers as ReadonlyArray<string>).includes(browserName);
+}
+export function validateBrowser(name: string): SupportedBrowser {
+    if (isValidBrowserName(name)) {
+        return name;
+    }
+    throw new Error(`Invalid browser name was entered as env var: ${name} \n
```
thehulke, line 11:
```suggestion
    throw new Error(`Invalid browser name was entered as env var: ${name} ,
```

AviVahl (approved):
fix conflicts

PR #1628, opened 2023-01-29
feature!: withFeature wait for load event by default

daomry:


packages/test-kit/src/with-feature.ts
```diff
@@ -337,6 +337,9 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
 
             const featurePage = await browserContext.newPage();
 
+            navigationOptions = navigationOptions || {};
+            navigationOptions.waitUntil = navigationOptions.waitUntil || 'load';
```
eladdo92, line 341: the default of `waitUntil` is 'load', maybe 'domcontentloaded'?
https://playwright.dev/docs/api/class-page#page-goto

PR #1645, opened 2023-02-14
tool(api): expose normalize-test-name function

shakederin:
Export the method that normalizes our test name to a valid file name.

Can be useful in trace files 

https://github.com/wixplosives/codux/pull/12983#:~:text=Can%20this%20logic%20be%20consumed%20from%20the%20engine%20(test%20kit)%3F


packages/test-kit/src/with-feature.ts
```diff
@@ -299,7 +301,7 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
                             name:
                                 process.env.TRACING && process.env.TRACING !== 'true'
                                     ? process.env.TRACING
-                                    : name ?? testName?.replace(/(\W+)/gi, ' ').trim().replace(/(\W+)/gi, '-'),
+                                    : name ?? (testName ? normalizeTestName(testName) : 'nameless-test'),
```
YardenPorat, line 304: Not sure about it, probably should keep an eye if there are occurrences of `nameless-test`

YardenPorat (approved):

packages/test-kit/src/normalize-test-name.ts
```diff
@@ -0,0 +1,3 @@
+export const normalizeTestName = (testName: string) => {
+    return testName.replace(/(\W+)/gi, ' ').trim().replace(/(\W+)/gi, '-');
+};
```
idango10, line 3:
```suggestion
export const normalizeTestName = (testName: string) => testName.replace(/(\W+)/gi, ' ').trim().replace(/(\W+)/gi, '-');
```

PR #1659, opened 2023-02-20
feat: expose environment disconnect details

olehrakwix:
Expose environment disconnect details such as exit code, signal, and last seen error to allow engine users to handle environment crash correctly in case of an unhandled exception.

### DIFF
--- a/packages/electron-commons/src/expirable-list.ts
+++ b/packages/electron-commons/src/expirable-list.ts
@@ -1,5 +1,5 @@
 /**
- * The collection of items that expires items on the specified expiration timeout.
+ * A list where items are automatically removed after a certain timeout.
  */
 export class ExpirableList<T> {
     private items: { item: T; timestamp: number }[] = [];
--- a/packages/electron-commons/src/expirable-list.ts
+++ b/packages/electron-commons/src/expirable-list.ts
@@ -1,0 +1,37 @@
+/**
+ * The collection of items that expires items on the specified expiration timeout.
+ */
+export class ExpirableList<T> {
+    private items: { item: T; timestamp: number }[] = [];
+
+    /**
+     * @param expirationTimeMilliseconds the time to live for items in milliseconds
+     */
+    constructor(private expirationTimeMilliseconds: number) {}
+
+    /**
+     * Adds new item to a collection
+     */
+    public push(item: T): void {
+        this.removeExpiredItems();
+        this.items.push({
+            item,
+            timestamp: performance.now(),
+        });
+    }
+
+    /**
+     * @returns the list of non-expired items
+     */
+    public getItems(): T[] {
+        this.removeExpiredItems();
+        return this.items.map(({ item }) => item);
+    }
+
+    private removeExpiredItems() {
+        const currentTimestamp = performance.now();
+        this.items = this.items.filter(
+            ({ timestamp }) => timestamp + this.expirationTimeMilliseconds > currentTimestamp
+        );
+    }
+}
--- a/packages/electron-commons/src/initializers/electron-node.ts
+++ b/packages/electron-commons/src/initializers/electron-node.ts
@@ -12,6 +12,16 @@ export interface InitializeNodeEnvironmentOptions extends InitializerOptions {
     processOptions?: Pick<SpawnOptions, 'cwd' | 'shell' | 'env' | 'stdio' | 'windowsHide'>;
 }
 
+export interface DisconnectHandler {
+    (details: DisconnectDetails): void;
+}
+
+export interface DisconnectDetails {
+    exitCode: number | null;
+    signal: NodeJS.Signals | null;
+    lastSeenError: string | undefined;
+}
+
 /**
  * Spawn a node-based environment from the renderer process.
  * Should be invoked *only* from the renderer process.
@@ -19,7 +29,12 @@ export interface InitializeNodeEnvironmentOptions extends InitializerOptions {
  */
 
 export const initializeNodeEnvironment: EnvironmentInitializer<
-    { id: string; dispose: () => void; onDisconnect: (cb: () => void) => void; environmentIsReady: Promise<void> },
+    {
+        id: string;
+        dispose: () => void;
+        onDisconnect: (cb: DisconnectHandler) => void;
+        environmentIsReady: Promise<void>;
+    },
     InitializeNodeEnvironmentOptions
 > = ({ communication, env, runtimeArguments, processOptions, environmentStartupOptions }) => {
     const environmentIsReady = communication.envReady(env.env);
@@ -60,7 +75,12 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
     child.stdout?.setEncoding('utf8');
     child.stderr?.setEncoding('utf8');
     child.stdout?.on('data', console.log); // eslint-disable-line no-console
-    child.stderr?.on('data', console.error); // eslint-disable-line no-console
+
+    let lastSeenError: string | undefined = undefined;
+    child.stderr?.on('data', (chunk) => {
+        console.error(chunk); // eslint-disable-line no-console
+        lastSeenError = chunk;
+    });
 
     return {
         id: env.env,
@@ -70,8 +90,10 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
                 child.pid ? await promisifiedTreeKill(child.pid) : child.kill();
             }
         },
-        onDisconnect: (cb: () => void) => {
-            child.on('exit', cb);
+        onDisconnect: (cb: DisconnectHandler) => {
+            child.on('exit', (exitCode, signal) => {
+                cb({ exitCode, signal, lastSeenError });
+            });
         },
         environmentIsReady,
     };
--- a/packages/electron-commons/src/initializers/electron-node.ts
+++ b/packages/electron-commons/src/initializers/electron-node.ts
@@ -84,7 +84,7 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
     child.stderr?.setEncoding('utf8');
     child.stdout?.on('data', console.log); // eslint-disable-line no-console
 
-    const errorChunks = new ExpirableList<string>(5 * 1000);
+    const errorChunks = new ExpirableList<string>(5_000);
 
     child.stderr?.on('data', (chunk) => {
         console.error(chunk); // eslint-disable-line no-console
--- a/packages/electron-commons/src/initializers/electron-node.ts
+++ b/packages/electron-commons/src/initializers/electron-node.ts
@@ -4,6 +4,7 @@ import type { EnvironmentInitializer, InitializerOptions } from '@wixc3/engine-c
 import type { IEngineRuntimeArguments, INodeEnvStartupMessage, NodeEnvironmentStartupOptions } from '../types';
 import treeKill from 'tree-kill';
 import { promisify } from 'util';
+import { ExpirableList } from '../expirable-list';
 const promisifiedTreeKill = promisify(treeKill);
 
 export interface InitializeNodeEnvironmentOptions extends InitializerOptions {
@@ -12,9 +13,6 @@ export interface InitializeNodeEnvironmentOptions extends InitializerOptions {
     processOptions?: Pick<SpawnOptions, 'cwd' | 'shell' | 'env' | 'stdio' | 'windowsHide'>;
 }
 
-/**
- * Detailed info about process exit.
- */
 export interface ProcessExitDetails {
     /**
      * The exit code if the child exited on its own
@@ -29,7 +27,7 @@ export interface ProcessExitDetails {
     /**
      * The last output process sent to stderr stream
      */
-    lastSeenError: string | null;
+    errorMessage: string;
 }
 
 /**
@@ -42,7 +40,7 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
     {
         id: string;
         dispose: () => void;
-        onDisconnect: (cb: (details: ProcessExitDetails) => void) => void;
+        onExit: (cb: (details: ProcessExitDetails) => void) => void;
         environmentIsReady: Promise<void>;
     },
     InitializeNodeEnvironmentOptions
@@ -86,22 +84,11 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
     child.stderr?.setEncoding('utf8');
     child.stdout?.on('data', console.log); // eslint-disable-line no-console
 
-    let lastErrors: string[] = [];
-    let lastErrorDateStamp: number | undefined = undefined;
+    const errorChunks = new ExpirableList<string>(5 * 1000);
 
     child.stderr?.on('data', (chunk) => {
         console.error(chunk); // eslint-disable-line no-console
-
-        if (lastErrorDateStamp === undefined) {
-            lastErrorDateStamp = Date.now();
-        }
-
-        // collect all errors if difference between occurrences is less than some threshold
-        if (lastErrorDateStamp - Date.now() < 5 * 1000) {
-            lastErrors.push(chunk);
-        } else {
-            lastErrors = [chunk];
-        }
+        errorChunks.push(chunk);
     });
 
     return {
@@ -112,9 +99,9 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
                 child.pid ? await promisifiedTreeKill(child.pid) : child.kill();
             }
         },
-        onDisconnect: (cb: (details: ProcessExitDetails) => void) => {
+        onExit: (cb: (details: ProcessExitDetails) => void) => {
             child.once('exit', (exitCode, signal) => {
-                cb({ exitCode, signal, lastSeenError: lastErrors.length > 0 ? lastErrors.join('') : null });
+                cb({ exitCode, signal, errorMessage: errorChunks.getItems().join('') });
             });
         },
         environmentIsReady,
--- a/packages/electron-commons/src/initializers/electron-node.ts
+++ b/packages/electron-commons/src/initializers/electron-node.ts
@@ -29,7 +29,7 @@ export interface ProcessExitDetails {
     /**
      * The last output process sent to stderr stream
      */
-    lastSeenError: string | undefined;
+    lastSeenError: string | null;
 }
 
 /**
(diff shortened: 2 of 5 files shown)


packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -12,14 +12,29 @@ export interface InitializeNodeEnvironmentOptions extends InitializerOptions {
     processOptions?: Pick<SpawnOptions, 'cwd' | 'shell' | 'env' | 'stdio' | 'windowsHide'>;
 }
 
+export interface DisconnectHandler {
+    (details: DisconnectDetails): void;
+}
+
+export interface DisconnectDetails {
```
alisey, line 19: I'd rather call it `ProcessExitDetails` because 1) it's about process, and 2) it's about exiting.

`disconnect` and `exit` both exist, but they're different events:
* https://nodejs.org/api/process.html#event-disconnect
* https://nodejs.org/api/process.html#event-exit

I would also remove `DisconnectHandler`, and instead use `(details: ProcessExitDetails) => void` inline to make it more self-descriptive, it wouldn't be too much boilerplate.

alisey (approved):

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -60,7 +75,12 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
     child.stdout?.setEncoding('utf8');
     child.stderr?.setEncoding('utf8');
     child.stdout?.on('data', console.log); // eslint-disable-line no-console
-    child.stderr?.on('data', console.error); // eslint-disable-line no-console
+
+    let lastSeenError: string | undefined = undefined;
+    child.stderr?.on('data', (chunk) => {
+        console.error(chunk); // eslint-disable-line no-console
+        lastSeenError = chunk;
```
alisey, line 82: In the case of a process crash (e.g. out of memoty) the entire output might not fit in a single chunk. Can you please verify?

Maybe we could include the last second of stderr output instead?

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -70,8 +90,10 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
                 child.pid ? await promisifiedTreeKill(child.pid) : child.kill();
             }
         },
-        onDisconnect: (cb: () => void) => {
-            child.on('exit', cb);
+        onDisconnect: (cb: DisconnectHandler) => {
+            child.on('exit', (exitCode, signal) => {
```
alisey, line 94: `child.once('exit', ...)`

packages/electron-commons/test/initializers/electron-node.spec.ts
```diff
@@ -54,21 +56,37 @@ describe('onDisconnectHandler for node environment initializer', () => {
...
+            });
         });
         it('should catch on env unhandled promise rejection', async () => {
             const { disconnectPromise } = await setupRunningEnv({ errorMode: 'promiseReject' });
 
-            await expect(disconnectPromise).to.eventually.eq(true);
+            await expect(disconnectPromise).to.eventually.be.fulfilled;
+        });
+        it('should expose error when env throwing uncaught exception', async () => {
+            const { disconnectPromise } = await setupRunningEnv({ errorMode: 'exception', pipeOutput: true });
```
alisey, line 85: I don't understand the thing with `pipeOutput`, shouldn't it work without it?

packages/electron-commons/test/initializers/electron-node.spec.ts
```diff
@@ -28,16 +29,17 @@ const setupRunningEnv = async ({
...
         environmentStartupOptions: {},
     });
 
     await environmentIsReady;
 
-    const disconnectPromise = new Promise<boolean>((res) => {
-        onDisconnect(() => {
-            res(true);
-        });
+    const disconnectPromise = new Promise<DisconnectDetails>((res) => {
```
AviVahl, line 41: the mixture of the api accepting a callback and us wanting to use promises in async flows is really shown here.

there are several ways we could alleviate this:
- expose the child process to outside to listen to exit on its own terms.
- use node's event emitter for these events with tests able to call `{once} from 'events'`.
- error handling, should be done via IPC and exposed as on-going error events every time an error occurs inside the child process

AviVahl (changes requested):
note about the getter assertion is true for all tests

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: thanks, renamed

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: You were right, I changed the logic to collect all errors if the time difference between them is less than 5 seconds.

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: fixed

packages/electron-commons/test/initializers/electron-node.spec.ts
olehrakwix: by default stdio was ignored in these tests completely. in order to check error exposing, I done this change. anyway, refactored this

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -12,14 +12,29 @@ export interface InitializeNodeEnvironmentOptions extends InitializerOptions {
     processOptions?: Pick<SpawnOptions, 'cwd' | 'shell' | 'env' | 'stdio' | 'windowsHide'>;
 }
 
+export interface DisconnectHandler {
```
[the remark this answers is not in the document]
olehrakwix: added

packages/electron-commons/src/initializers/electron-node.ts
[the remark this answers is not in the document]
olehrakwix: added

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -60,7 +75,12 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
     child.stdout?.setEncoding('utf8');
     child.stderr?.setEncoding('utf8');
     child.stdout?.on('data', console.log); // eslint-disable-line no-console
-    child.stderr?.on('data', console.error); // eslint-disable-line no-console
+
+    let lastSeenError: string | undefined = undefined;
+    child.stderr?.on('data', (chunk) => {
```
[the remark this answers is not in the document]
olehrakwix: it is true, however, it is better to collect and expose errors anyway.
saying that we should catch all errors is like saying 'new bug' button is not needed in jira, cause we will not create bugs 😀

packages/electron-commons/src/initializers/electron-node.ts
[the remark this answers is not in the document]
olehrakwix: updated to once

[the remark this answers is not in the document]
olehrakwix: this is not about promise even. we subscribing to child.exit event and exposing it as onDisconnect event for initialized environment

packages/electron-commons/test/initializers/electron-node.spec.ts
```diff
@@ -28,16 +29,17 @@ const setupRunningEnv = async ({
             config: [testFeature.use({ errorType: { type: errorMode, handleUncaught } })],
             features: Array.from(features.entries()),
         },
-        processOptions: { cwd: process.cwd(), stdio: ['ignore', 'ignore', 'ignore', 'ipc'] },
+        processOptions: {
+            cwd: process.cwd(),
+            stdio: pipeOutput === true ? ['pipe', 'pipe', 'pipe', 'ipc'] : ['ignore', 'ignore', 'ignore', 'ipc'],
```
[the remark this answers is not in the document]
olehrakwix: thanks, really make sense, refactored this

packages/electron-commons/test/initializers/electron-node.spec.ts
```diff
@@ -54,21 +56,37 @@ describe('onDisconnectHandler for node environment initializer', () => {
 
             dispose();
 
-            await expect(disconnectPromise).to.eventually.eq(true);
+            await expect(disconnectPromise).to.eventually.be.fulfilled;
```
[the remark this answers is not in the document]
olehrakwix: maybe yes, maybe no, anyway, fixed to validate value

packages/electron-commons/test/initializers/electron-node.spec.ts
olehrakwix: This is just about how we are waiting process exit to test it. I'd not put too much attention to this.
also, please note that I haven't introduced any new approach here. this particular thing was like this for a long time. In this PR I've added just process exit details that has no relation to promises here.
however, changed this to deferred for more readability

olehrakwix:
@AviVahl , what do you mean by "note about the getter assertion is true for all tests"?

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -12,14 +12,39 @@ export interface InitializeNodeEnvironmentOptions extends InitializerOptions {
     processOptions?: Pick<SpawnOptions, 'cwd' | 'shell' | 'env' | 'stdio' | 'windowsHide'>;
 }
 
+/**
+ * Detailed info about process exit.
+ */
+export interface ProcessExitDetails {
+    /**
+     * The exit code if the child exited on its own
```
alisey, line 20: The comment makes me question what "on its own" means. Not having a comment would make this absolutely clear.

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -12,14 +12,39 @@ export interface InitializeNodeEnvironmentOptions extends InitializerOptions {
...
+     */
+    exitCode: number | null;
+
+    /**
+     * The signal by which the child process was terminated
```
alisey, line 25: If the process traps the signals and exits, is it terminated by the signal?

Like with the previous property, not having a description would give me more clarity.

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -12,14 +12,39 @@ export interface InitializeNodeEnvironmentOptions extends InitializerOptions {
     processOptions?: Pick<SpawnOptions, 'cwd' | 'shell' | 'env' | 'stdio' | 'windowsHide'>;
 }
 
+/**
+ * Detailed info about process exit.
+ */
```
alisey, line 17: I don't think this comment tells me anything. I'd rather not have it.

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -12,14 +12,39 @@ export interface InitializeNodeEnvironmentOptions extends InitializerOptions {
...
  * Should be invoked *only* from the renderer process.
  * The caller should await returned `environmentIsReady` promise to ensure environment initialization is finished.
  */
 
 export const initializeNodeEnvironment: EnvironmentInitializer<
-    { id: string; dispose: () => void; onDisconnect: (cb: () => void) => void; environmentIsReady: Promise<void> },
+    {
+        id: string;
+        dispose: () => void;
+        onDisconnect: (cb: (details: ProcessExitDetails) => void) => void;
```
alisey, line 45: onExit

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -60,7 +85,24 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
     child.stdout?.setEncoding('utf8');
     child.stderr?.setEncoding('utf8');
     child.stdout?.on('data', console.log); // eslint-disable-line no-console
-    child.stderr?.on('data', console.error); // eslint-disable-line no-console
+
+    let lastErrors: string[] = [];
+    let lastErrorDateStamp: number | undefined = undefined;
```
alisey, line 90: DateStamp - it's called timestamp, a single word.

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -60,7 +85,24 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
...
+
+    child.stderr?.on('data', (chunk) => {
+        console.error(chunk); // eslint-disable-line no-console
+
+        if (lastErrorDateStamp === undefined) {
+            lastErrorDateStamp = Date.now();
+        }
+
+        // collect all errors if difference between occurrences is less than some threshold
+        if (lastErrorDateStamp - Date.now() < 5 * 1000) {
```
alisey, line 100: Never use `Date.now()` to measure time intervals. You could even get a negative interval.

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -60,7 +85,24 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
...
+    let lastErrorDateStamp: number | undefined = undefined;
+
+    child.stderr?.on('data', (chunk) => {
+        console.error(chunk); // eslint-disable-line no-console
+
+        if (lastErrorDateStamp === undefined) {
+            lastErrorDateStamp = Date.now();
+        }
+
+        // collect all errors if difference between occurrences is less than some threshold
```
alisey, line 99: This comment is not helping. We don't collect all errors.

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -60,7 +85,24 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
...
+
+    let lastErrors: string[] = [];
+    let lastErrorDateStamp: number | undefined = undefined;
+
+    child.stderr?.on('data', (chunk) => {
+        console.error(chunk); // eslint-disable-line no-console
+
+        if (lastErrorDateStamp === undefined) {
+            lastErrorDateStamp = Date.now();
+        }
```
alisey, line 97: This basically makes it the the first error timestamp, not the last.

Instead add at the end of the handler:

```ts
lastErrorTs = performance.now();
```

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -60,7 +85,24 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
...
+
+        // collect all errors if difference between occurrences is less than some threshold
+        if (lastErrorDateStamp - Date.now() < 5 * 1000) {
+            lastErrors.push(chunk);
```
alisey, line 101: If the process periodically writes to stderr, then simply comparing timestamp of the last event could result in all of the output being included.

On one hand, this is not very likely, on the other hand, changing the code to collect only the last five seconds is trivial.

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -12,14 +12,39 @@ export interface InitializeNodeEnvironmentOptions extends InitializerOptions {
...
+
+    /**
+     * The signal by which the child process was terminated
+     */
+    signal: NodeJS.Signals | null;
+
+    /**
+     * The last output process sent to stderr stream
+     */
+    lastSeenError: string | null;
```
alisey, line 32: Isn't it the same as `lastError`? What does `seen` change?

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: this is taken from node documentation about child exit event

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: this is taken from node documentation about child exit event

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: renamed

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: renamed to timestamp and refactored

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: fixed this

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: removed

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: refactored this

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: refactored this

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: renamed

packages/electron-commons/src/expirable-list.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * The collection of items that expires items on the specified expiration timeout.
```
alisey, line 2: A list where items are automatically removed after a certain timeout.

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -60,7 +83,13 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
     child.stdout?.setEncoding('utf8');
     child.stderr?.setEncoding('utf8');
     child.stdout?.on('data', console.log); // eslint-disable-line no-console
-    child.stderr?.on('data', console.error); // eslint-disable-line no-console
+
+    const errorChunks = new ExpirableList<string>(5 * 1000);
```
alisey, line 87: `5_000` 🙃

packages/electron-commons/src/expirable-list.ts
olehrakwix: done

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: done

alisey (approved):

AviVahl (approved):

PR #1712, opened 2023-03-29
chore(createEngineHttpServer): added hostname param

thehulke:
This allows us to provide `hostname` param to newly created http server.

`hostname` option added to `launchEngineHttpServer` method.

### DIFF
--- a/packages/runtime-node/src/launch-http-server.ts
+++ b/packages/runtime-node/src/launch-http-server.ts
@@ -21,6 +21,7 @@ export interface ILaunchHttpServerOptions {
     httpServerPort?: number;
     socketServerOptions?: Partial<io.ServerOptions>;
     routeMiddlewares?: Array<RouteMiddleware>;
+    hostname?: string;
 }
 
 export async function launchEngineHttpServer({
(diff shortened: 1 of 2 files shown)


packages/runtime-node/src/launch-http-server.ts
```diff
@@ -21,20 +21,22 @@ export interface ILaunchHttpServerOptions {
...
     routeMiddlewares = [],
+    hostname,
 }: ILaunchHttpServerOptions = {}) {
     const app = express();
     for (const { path, handlers } of routeMiddlewares) {
         app.use(path, handlers);
     }
     app.use(cors());
-    const { port, httpServer } = await safeListeningHttpServer(httpServerPort, app);
+    const { port, httpServer } = await safeListeningHttpServer(httpServerPort, app, 100, hostname);
```
olehrakwix, line 39: is it better to pass `undefined` instead of 100 to force the function to use its default value?

olehrakwix (approved):

AviVahl (approved):

AviVahl:
not sure why passing in an hostname from devserver fails tests on windows/mac. you may need to check behavior there...

barak007:
outdated

PR #1716, opened 2023-03-30
chore!: moved types between packages to be able to use them in future PR

olehrakwix:


AviVahl (approved):

packages/runtime-node/src/types.ts
```diff
@@ -134,3 +134,56 @@ export interface IConfigDefinition {
...
+    packageName: string;
+    filePath: string;
+    exportedEnvs?: IEnvironmentDescriptor[];
+}
+
+export const metadataApiToken = {
+    id: 'metadata-api-token',
+};
+
+export const isWorkerThreadEnvStartupMessage = (value: unknown): value is IWorkerThreadEnvStartupMessage => {
```
yurii-ve, line 171: I can't find a place where this type was moved from.. is it a new type?

olehrakwix: this is new one. not directly related to this particular refactoring, but does not break anything

PR #1717, opened 2023-03-30
fix!: removed limitation for contextual env to be single

olehrakwix:


alisey (approved):
I can't claim I understand it, but I like it.

cylon3035 (approved):
Amazing ✨
lets just not forget that this is braking change and we need to bump major

PR #1718, opened 2023-03-30
refactor: improved test, make feature init ready for reuse

olehrakwix:
- make `setupRunningNodeEnv` more generic and move it from test to testkit
- fixed `setupRunningNodeEnv` to run real feature discovery instead of having it hardcoded
- prepare `setupRunningNodeEnv` for reuse
- moved `test-project` from package to `test-fixtures`


cylon3035 (approved):

packages/electron-commons/test/initializers/electron-node.spec.ts
```diff
@@ -1,60 +1,31 @@
...
 
-const setupRunningEnv = async ({
-    errorMode,
-    handleUncaught,
-    stdio = 'ignore',
-}: {
-    errorMode?: 'exception' | 'exit' | 'promiseReject' | 'out-of-memory' | 'no-error';
-    handleUncaught?: boolean;
+interface SetupRunningFeatureOptions {
+    config: Partial<ErrorTypeConfig>;
```
alisey, line 16: Options object that contains a config object? Please no.

olehrakwix: `options` - the options of environment initialization
`config` - config of feature that is being initialized

alisey: Then call it featureConfig. Isn't it weird that feature config has type `ErrorTypeConfig`?

alisey: Also, `errorMode` in feature config? Are we taking about the same feature config?
https://github.com/wixplosives/engine/blob/master/packages/core/src/entities/config.ts#L9

PR #1719, opened 2023-03-30
feat: added workerthread environment type support

olehrakwix:
- `workerthread` env type
- `workerthread` env initializer
- `workerthread` entrypoint
- tests to verify we can create new worker env for both `single` and `multi` modes and call its' API

### DIFF
--- a/packages/core/src/com/types.ts
+++ b/packages/core/src/com/types.ts
@@ -7,6 +7,7 @@ export type EnvironmentTypes =
     | 'window'
     | 'iframe'
     | 'webworker'
+    | 'workerthread'
     | 'node'
     | 'context'
     | 'electron-renderer'
--- a/packages/electron-commons/test/worker-thread.spec.ts
+++ b/packages/electron-commons/test/worker-thread.spec.ts
@@ -1,0 +1,35 @@
+import fs from '@file-services/node';
+import chai, { expect } from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+
+import workerThreadFeature, { serverEnv } from '@fixture/worker-thread/dist/worker-thread.feature';
+
+import { setupRunningNodeEnv } from '../test-kit/setup-running-node-env';
+
+chai.use(chaiAsPromised);
+
+const featurePath = fs.dirname(require.resolve('@fixture/worker-thread/package.json'));
+
+const setupRunningEnv = (featureId: string) =>
+    setupRunningNodeEnv({
+        featurePath,
+        featureId,
+        env: serverEnv,
+    });
+
+describe('workerthread environment type', () => {
+    it('initializes and exposes API', async () => {
+        const { exitPromise } = await setupRunningEnv(workerThreadFeature.id);
+
+        const exitResult = await exitPromise;
+
+        expect(exitResult.exitCode).to.eql(0);
+    });
+    it('initializes multi workerthread environment and exposes API', async () => {
+        const { exitPromise } = await setupRunningEnv(`${workerThreadFeature.id}/multi`);
+
+        const exitResult = await exitPromise;
+
+        expect(exitResult.exitCode).to.eql(0);
+    });
+});
--- a/packages/runtime-node/package.json
+++ b/packages/runtime-node/package.json
@@ -21,7 +21,8 @@
     "!dist/test",
     "src",
     "!*/tsconfig.{json,tsbuildinfo}",
-    "templates"
+    "templates",
+    "worker-thread-entry.js"
   ],
   "license": "MIT",
   "publishConfig": {
--- a/packages/runtime-node/src/worker-thread-entry.ts
+++ b/packages/runtime-node/src/worker-thread-entry.ts
@@ -1,0 +1,63 @@
+import { COM } from '@wixc3/engine-core';
+import { parentPort } from 'node:worker_threads';
+import { importModules } from './import-modules';
+
+import { runNodeEnvironment } from './node-environment';
+import { isWorkerThreadEnvStartupMessage } from './types';
+import { WorkerThreadHost } from './worker-thread-host';
+
+const messageHandler = async (message: unknown) => {
+    if (isWorkerThreadEnvStartupMessage(message)) {
+        if (parentPort === null) {
+            throw new Error('this file should be executed in `worker_thread` context');
+        }
+
+        const {
+            requiredModules,
+            basePath,
+            environmentName,
+            config,
+            environmentContextName,
+            featureName,
+            features,
+            parentEnvName,
+            env,
+        } = message.runOptions;
+        if (requiredModules) {
+            await importModules(basePath, requiredModules);
+        }
+
+        const host = new WorkerThreadHost(parentPort);
+
+        config.push(
+            COM.use({
+                config: {
+                    connectedEnvironments: {
+                        [parentEnvName]: {
+                            id: parentEnvName,
+                            host,
+                        },
+                    },
+                },
+            })
+        );
+
+        await runNodeEnvironment({
+            env,
+            featureName,
+            features,
+            config,
+            host,
+            name: environmentName,
+            type: 'workerthread',
+            childEnvName: environmentContextName,
+        });
+    }
+};
+
+if (parentPort !== null) {
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
+    parentPort.once('message', messageHandler);
+} else {
+    throw new Error('this file should be executed in `worker_thread` context');
+}
--- a/packages/runtime-node/src/worker-thread-host.ts
+++ b/packages/runtime-node/src/worker-thread-host.ts
@@ -1,0 +1,33 @@
+import { MessagePort, Worker } from 'node:worker_threads';
+
+import { BaseHost, Message } from '@wixc3/engine-core';
+
+export class WorkerThreadHost extends BaseHost {
+    constructor(private host: Worker | MessagePort) {
+        super();
+        host.on('message', this.onMessage);
+    }
+
+    public addEventListener(type: 'message', handler: (event: { data: any }) => void): void {
+        const handlers = this.handlers.get(type);
+        if (!handlers) {
+            this.handlers.set(type, new Set([handler]));
+        } else {
+            handlers.add(handler);
+        }
+
+        this.host.addListener(type, handler);
+    }
+
+    public removeEventListener(type: 'message', handler: (event: { data: any }) => void): void {
+        this.host.removeListener(type, handler);
+    }
+
+    public postMessage(data: Message): void {
+        this.host.postMessage(data);
+    }
+
+    private onMessage: (...args: any[]) => void = (message) => {
+        this.emitMessageHandlers(message);
+    };
+}
--- a/packages/runtime-node/src/worker-thread-initializer.ts
+++ b/packages/runtime-node/src/worker-thread-initializer.ts
@@ -1,0 +1,58 @@
+import { Worker } from 'node:worker_threads';
+
+import { COM, InitializerOptions } from '@wixc3/engine-core';
+
+import { WorkerThreadHost } from './worker-thread-host';
+import type { IWorkerThreadEnvStartupMessage } from './types';
+import { getApplicationMetaData } from './get-application-metadata';
+
+export async function workerThreadInitializer({ communication, env }: InitializerOptions) {
+    const isSingleton = env.endpointType === 'single';
+    const instanceId = isSingleton ? env.env : communication.generateEnvInstanceID(env.env);
+    const envReady = communication.envReady(instanceId);
+
+    const { workerThreadEntryPath, requiredModules, basePath, config, featureName, features } =
+        await getApplicationMetaData(communication);
+
+    const worker = new Worker(workerThreadEntryPath, {
+        workerData: {
+            name: instanceId,
+        },
+    });
+
+    const host = new WorkerThreadHost(worker);
+    communication.registerEnv(instanceId, host);
+    communication.registerMessageHandler(host);
+
+    const runOptions = {
+        requiredModules,
+        basePath,
+        environmentName: instanceId,
+        /**
+         * configuration contains data that can not be serialized to worker communication channel.
+         * in the configuration of COM feature there is LOCAL_ENVIRONMENT_INITIALIZER_ENV_ID
+         * environment info that is needed when node env wants to start new node env.
+         * this is not a case for workerthread, so ignoring that config here
+         */
+        config: config.filter(([featureId]) => featureId !== COM.id),
+        featureName,
+        features,
+        parentEnvName: communication.getEnvironmentName(),
+        env,
+    };
+
+    worker.postMessage({
+        id: 'workerThreadStartupOptions',
+        runOptions,
+    } as IWorkerThreadEnvStartupMessage);
+
+    await envReady;
+
+    return {
+        id: instanceId,
+        dispose: async () => {
+            await worker.terminate();
+            communication.clearEnvironment(instanceId);
+        },
+    };
+}
--- a/test-fixtures/worker-thread/package.json
+++ b/test-fixtures/worker-thread/package.json
@@ -1,0 +1,16 @@
+{
+  "name": "@fixture/worker-thread",
+  "version": "1.0.0",
+  "main": "dist/entry.js",
+  "scripts": {},
+  "dependencies": {
+    "@wixc3/engine-core": "^33.2.0",
+    "@wixc3/engine-runtime-node": "^33.2.0"
+  },
+  "license": "unlicensed",
+  "publishConfig": {
+    "access": "restricted"
+  },
+  "private": true,
+  "author": "Wix.com"
+}
--- a/test-fixtures/worker-thread/src/multi.server.env.ts
+++ b/test-fixtures/worker-thread/src/multi.server.env.ts
@@ -1,0 +1,33 @@
+import { workerThreadInitializer } from '@wixc3/engine-runtime-node';
+import multiFeature, { serverEnv, workerEnv } from './multi.feature';
+
+const workersCount = 3;
+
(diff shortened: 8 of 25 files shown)


packages/electron-commons/test/worker-thread.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+        expect(exitResult.exitCode).to.eql(0);
+    });
+    it('initializes multi workerthread environment and exposes API', async () => {
+        const { exitPromise } = await setupRunningEnv(`${workerThreadFeature.id}/multi`);
+
+        const exitResult = await exitPromise;
+
+        expect(exitResult.exitCode).to.eql(0);
+    });
+    xit('initializes multi workerthread environment and exposes API if called from async method', async () => {
```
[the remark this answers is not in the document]
olehrakwix: thanks, nice catch. this test was added when I found the issue, but now it is fixed in separate PR with own test

yarn.lock
```diff
@@ -5635,6 +5635,11 @@ memfs@^3.4.12:
   dependencies:
     fs-monkey "^1.0.3"
 
+memoize-one@^6.0.0:
```
barak007, line 5638: seems like it's not in use

packages/runtime-node/package.json
```diff
@@ -13,6 +13,7 @@
     "@wixc3/engine-core-node": "^33.2.0",
     "@wixc3/patterns": "^8.1.0",
     "create-listening-server": "^2.1.0",
+    "memoize-one": "^6.0.0",
```
AviVahl, line 16: this is unused. remove

packages/runtime-node/src/worker-thread-entry.ts
```diff
@@ -0,0 +1,63 @@
...
+            host,
+            name: environmentName,
+            type: 'workerthread',
+            childEnvName: environmentContextName,
+        });
+    }
+};
+
+if (parentPort !== null) {
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
AviVahl, line 59: what handles rejections if runNodeEnvironment fails?

packages/runtime-node/src/worker-thread-initializer.ts
```diff
@@ -0,0 +1,58 @@
...
+        id: 'workerThreadStartupOptions',
+        runOptions,
+    } as IWorkerThreadEnvStartupMessage);
+
+    await envReady;
+
+    return {
+        id: instanceId,
+        dispose: async () => {
+            await worker.terminate();
```
AviVahl, line 54: rather than trying to force terminate, don't we want to pass a close message so feature can cleanup after itself?

test-fixtures/worker-thread/src/multi.server.env.ts
```diff
@@ -0,0 +1,33 @@
...
+            if (response !== workerResponseConfig.response) {
+                exitCode = 1;
+            }
+        }
+
+        for (const { dispose } of workerEnvs) {
+            await dispose();
+        }
+
+        process.exit(exitCode);
```
AviVahl, line 31: this forces the process to close and is a bad practice

test-fixtures/worker-thread/src/worker-thread.server.env.ts
```diff
@@ -0,0 +1,21 @@
...
+            communication,
+            env: workerEnv,
+        });
+
+        const response = await workerEcho.ping();
+
+        await disposeWorker();
+
+        if (response === workerResponseConfig.response) {
+            process.exit(0);
```
AviVahl, line 16: avoid this

test-fixtures/worker-thread/src/multi.server.env.ts
```diff
@@ -0,0 +1,33 @@
...
+            )
+        );
+
+        let exitCode = 0;
+        for (const workerEnv of workerEnvs) {
+            const workerEchoService = multiWorkerEcho.get({ id: workerEnv.id });
+            const response = await workerEchoService.ping();
+
+            if (response !== workerResponseConfig.response) {
+                exitCode = 1;
```
AviVahl, line 23: so if the next worker succeeds, this gets overridden?

AviVahl (changes requested):

yarn.lock
olehrakwix: removed

packages/core-node/src/types.ts
```diff
@@ -31,6 +31,7 @@ export interface IEngineRuntimeArguments {
     configName?: string;
     devport?: number;
     nodeEntryPath: string;
+    workerThreadEntryPath: string;
```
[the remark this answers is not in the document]
olehrakwix: no, we need both, nodeEntryPath is used to spawn new node environment and workerThreadEntryPath is used to spawn worker_thread

packages/core/src/com/types.ts
```diff
@@ -60,6 +61,10 @@ export type AsyncApi<T extends object> = {
         : never;
 };
 
+export type MultiEnvAsyncApi<T extends object> = {
```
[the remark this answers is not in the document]
olehrakwix: added

packages/runtime-node/package.json
olehrakwix: revmoed

test-fixtures/worker-thread/src/multi.server.env.ts
olehrakwix: I agree, tried other approaches and could not find better solution. added a separate test that checks if dispose is happening correctly, but left process.exit as it is here

test-fixtures/worker-thread/src/multi.server.env.ts
olehrakwix: no, this overrides with '1' only if response differs from expected one

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -96,8 +112,16 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
...
-        // eslint-disable-next-line @typescript-eslint/no-misused-promises
         dispose: async () => {
+            const envDisposed = executeRemoteCall<NodeEnvironmentDisposeCommand, NodeEnvironmentDisposedEvent>(
+                child,
+                { id: 'nodeEnvironmentDisposeCommand' },
+                'nodeEnvironmentDisposedEvent'
+            );
+
+            // wait for dispose finished or timeout
+            await Promise.allSettled([timeout(envDisposed, 1_000)]);
```
AviVahl, line 123: 1s timeout? why?

olehrakwix: to terminate process if dispose is taking a long time

packages/electron-commons/src/initializers/electron-node.ts
```diff
@@ -76,7 +82,17 @@ export const initializeNodeEnvironment: EnvironmentInitializer<
         }
     );
 
-    child.send({ id: 'nodeStartupOptions', runOptions: nodeEnvStartupArguments } as INodeEnvStartupMessage);
+    const envInitFailed = deferred();
```
AviVahl, line 85: - deferred promises are a REALLY good way to make code harder to read. we should avoid if possible
- promise itself is named envInitFailed, which is odd. this is also why you've Promise.raced at the bottom. it never resolves (even when init failed). instead it only rejects. the double negatives are confusing.
- having a promise for envReady AND envInitFailed is confusing. I'd expect **envReady** to reject if anything fails. not to have a separate promise that only conditionally fails

packages/electron-commons/src/node-entry.ts
```diff
@@ -1,78 +1,112 @@
...
+
+const messageHandler = (message: unknown) => {
+    const command = message as NodeEnvironmentCommand;
+
+    switch (command.id) {
+        case 'nodeEnvironmentStartupCommand':
+            handleStartupCommand(command).catch((e) => {
+                emitEvent<NodeEnvironmentEvent>(process, {
+                    id: 'nodeEnvironmentInitFailedEvent',
+                    error: toError(e).message,
```
AviVahl, line 93: this is a terrible practice. it looses the stack trace. avoid

packages/electron-commons/src/node-entry.ts
```diff
@@ -1,78 +1,112 @@
...
-            config,
-            features,
-            options: runtimeOptions,
-            context: basePath,
-            parentEnvName,
-            env,
-        });
+        case 'nodeEnvironmentDisposeCommand':
+            handleDisposeCommand()
+                .then(() => process.off('message', messageHandler))
```
AviVahl, line 100: why isn't this inside handleDisposeCommand?

packages/electron-commons/src/node-entry.ts
```diff
@@ -1,78 +1,112 @@
...
-            features,
-            bundlePath,
-            runtimeOptions,
-            parentEnvName,
-            env,
-        } = message.runOptions;
-        if (requiredModules) {
-            await importModules(basePath, requiredModules);
-        }
+let disposeEnv: (() => Promise<void>) | undefined;
```
AviVahl, line 13: switch to a Set that contains openEnvs
iterate over it on close
clear it once done

packages/electron-commons/test/initializers/electron-node.spec.ts
```diff
@@ -41,15 +44,31 @@ describe('onDisconnectHandler for node environment initializer', () => {
...
     describe('without own uncaughtException handling', () => {
         it('should catch on dispose of env', async () => {
             const { dispose, exitPromise } = await setupRunningEnv({ config: { errorMode: 'no-error' } });
 
-            dispose();
+            await dispose();
+            const result = await exitPromise;
 
-            await expect(exitPromise).to.eventually.deep.eq(expectedTerminationResult);
+            // oneOf is done to avoid flakiness when env is disposing long time and terminated
```
AviVahl, line 60: it's only flaky cuz of the weird 1s

packages/electron-commons/test-kit/assert-utils.ts
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Asserts that value is deep equal to one of the specified expected values.
+ * @param actual value to test
+ * @param list the list of expected values
+ */
+export function oneOfDeepEqual(actual: any, list: any[]) {
+    expect(actual).to.satisfy(() => {
+        for (const expected of list) {
+            try {
+                expect(actual).to.deep.equal(expected);
```
AviVahl, line 12: this function would throw for any expectation that isn't equal
also, the error message will not be descriptive

packages/electron-commons/test/initializers/electron-node.spec.ts
```diff
@@ -41,15 +44,31 @@ describe('onDisconnectHandler for node environment initializer', () => {
         errorMessage: '',
     };
 
+    const expectedDisposeResult: ProcessExitDetails = {
+        exitCode: 0,
+        signal: null,
+        errorMessage: '',
```
AviVahl, line 50: errorMessage should not be defined when there was no error
also, it should not default to an empty string. most stuff shouldn't.

packages/runtime-node/src/communication-helpers.ts
```diff
@@ -0,0 +1,65 @@
+import { ChildProcess, Serializable } from 'node:child_process';
+import { type MessagePort, type Worker } from 'node:worker_threads';
+import { deferred } from 'promise-assist';
+
+/**
+ * Emits an event using specified emitter as MessagePort
+ */
+export function emitEvent<T extends Serializable>(
```
AviVahl, line 8: this shouldn't be generic. it's an anti-pattern to use it in a single fn parameter.
instead, caller should use `satisfies` on value being passed to this fn
and it should accept `event: unknown`

packages/runtime-node/src/communication-helpers.ts
```diff
@@ -0,0 +1,65 @@
...
+        emitter.postMessage(command);
+    }
+    return response.promise;
+}
+
+/**
+ * Subscribes for an even using specified event emitter
+ */
+export function onEvent<T extends { id: string }>(emitter: Worker | ChildProcess, handler: (e: T) => void) {
+    emitter.on('message', (e) => {
```
AviVahl, line 59: bad pattern. nobody is able to clean this listener

packages/runtime-node/src/communication-helpers.ts
```diff
@@ -0,0 +1,65 @@
+import { ChildProcess, Serializable } from 'node:child_process';
+import { type MessagePort, type Worker } from 'node:worker_threads';
+import { deferred } from 'promise-assist';
+
+/**
+ * Emits an event using specified emitter as MessagePort
+ */
+export function emitEvent<T extends Serializable>(
+    emitter: MessagePort | Worker | NodeJS.Process | ChildProcess,
```
AviVahl, line 9: not sure we want to combine it all to a single fn.
some of these require json serializable data, while others don't.

test-fixtures/worker-thread/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@fixture/worker-thread",
+  "version": "33.2.0",
```
AviVahl, line 3: master at 34 as of yesterday I believe

AviVahl (changes requested):
see comments

packages/electron-commons/src/initializers/electron-node.ts
olehrakwix: removed handling of env init failing, it would be hard to make this in same promise, cause envIsReady promise is created in communication, but communication know nothing about env init failure...
there is no init fail handling in current version, so if we do need this, it is better to implement separately

packages/electron-commons/src/node-entry.ts
olehrakwix: fixed, passing whole error object in message

packages/electron-commons/src/node-entry.ts
olehrakwix: fixed

packages/electron-commons/test/initializers/electron-node.spec.ts
olehrakwix: removed that force kill think

packages/electron-commons/test-kit/assert-utils.ts
olehrakwix: removed this function

packages/electron-commons/test/initializers/electron-node.spec.ts
olehrakwix: done

packages/runtime-node/src/communication-helpers.ts
olehrakwix: removed these helpers

packages/runtime-node/src/communication-helpers.ts
olehrakwix: removed these helpers

packages/runtime-node/src/communication-helpers.ts
olehrakwix: removed these helpers

test-fixtures/worker-thread/package.json
olehrakwix: fixed

packages/electron-commons/test/worker-thread.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+    it('initializes multi workerthread environment and exposes API', async () => {
+        const { exitPromise } = await setupRunningEnv(`${workerThreadFeature.id}/${multiFeature.id}`);
+        await expect(exitPromise).to.eventually.have.property('exitCode').eql(0);
+    });
+
+    it('correctly disposes worker_thread env', async () => {
+        const { dispose } = await setupRunningEnv(`${workerThreadFeature.id}/${emptyFeature.id}`);
+
+        // wait for worker thread being initialized inside feature
+        await sleep(1000);
```
AviVahl, line 39: this feels like a race condition (flakiness) waiting to happen
if we have a way to await the readiness of initialized environments (inclding the worker), we should
I specifically wonder what would happen if someone calls dispose before initialization ended (e.g. closing window while things are still loading)

packages/runtime-node/src/worker-thread-initializer.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    return {
+        id: instanceId,
+        dispose: async () => {
+            return new Promise((resolve, reject) => {
+                const handleWorkerDisposed = (e: unknown) => {
+                    if ((e as WorkerThreadDisposedEvent).id === 'workerThreadDisposedEvent') {
+                        worker.off('message', handleWorkerDisposed);
+                    }
+                    worker.terminate().then(resolve).catch(reject);
```
AviVahl, line 59: do we really need to forcibly terminate the worker? once all listeners inside are cleared, it should close on its own

packages/runtime-node/src/worker-thread-host.ts
```diff
@@ -0,0 +1,33 @@
...
+export class WorkerThreadHost extends BaseHost {
+    constructor(private host: Worker | MessagePort) {
+        super();
+        host.on('message', this.onMessage);
+    }
+
+    public addEventListener(type: 'message', handler: (event: { data: any }) => void): void {
+        const handlers = this.handlers.get(type);
+        if (!handlers) {
+            this.handlers.set(type, new Set([handler]));
```
AviVahl, line 14: a nice future refactor is to use SetMultiMap for handlers in BaseHost to avoid managing it in every extending host. not for this PR though

AviVahl (approved):

packages/electron-commons/test/worker-thread.spec.ts
olehrakwix: finally found a way to do it correctly, see last update

packages/runtime-node/src/worker-thread-initializer.ts
olehrakwix: you are right, removed this

PR #1747, opened 2023-04-12
fix: support optional function parameters in cross-environment method calls

### LINKED ISSUE #1434: Services: 'undefined' in method arguments gets converted to null when calling a remote service
If you have the following service:
```ts
class EchoService {
    public echo(message = 'hello world') {
        return message;
    }
}
```

Calling it from a different environment as `EchoService.echo(undefined)` will return `null` instead of `'hello world'`.

This has tripped me a few times. Undefined should stay undefined.

vladyslav-baliuk:
Fixes #1434


AviVahl:
Use stringify/parser replacer+reviver parameters instead?

packages/core/src/com/communication.ts
```diff
@@ -240,13 +246,14 @@ export class Communication {
         envId: string,
         api: string,
         method: string,
-        args: unknown[],
+        unserializedArgs: unknown[],
```
alisey, line 249: Just `args`

packages/core/src/com/communication.ts
```diff
@@ -240,13 +246,14 @@ export class Communication {
...
         origin: string,
         serviceComConfig: Record<string, AnyServiceMethodOptions>,
         forwardingChain: string[]
     ): Promise<unknown> {
         return new Promise<void>((res, rej) => {
             const callbackId = !serviceComConfig[method]?.emitOnly ? this.idsCounter.next('c') : undefined;
+            const args = serializeApiCallArguments(unserializedArgs);
```
alisey, line 256: Do it only if it's not a listen call.

packages/core/src/com/helpers.ts
```diff
@@ -23,3 +23,11 @@ export class MultiCounter {
         return `${ns}${this.ids[ns]++}`;
     }
 }
+
+const undefinedArgPlaceholder = '__|UND_ARG_PLA|__';
```
alisey, line 27: undefinedPlaceholder = '__undefined_placeholder__';

packages/core/src/com/helpers.ts
```diff
@@ -23,3 +23,11 @@ export class MultiCounter {
...
+
+export const serializeApiCallArguments = (args: unknown[]): unknown[] =>
```
alisey, line 29: Add a comment or a link to the issue explaining why this is necessary.

alisey:
> fix: added serialization/deserialization for arguments of the API calls

This is a not a reasonable title for this pull request.

packages/core/test/node/com.spec.ts
```diff
@@ -160,6 +160,25 @@ describe('Communication', () => {
         expect(handleMessageStub).to.have.not.been.called;
     });
 
+    it('serialization of arguments for remote API calls should work as expected ', async () => {
```
olehrakwix, line 163: "should work as expected" does not really help to understand what is expected

test-fixtures/simple-feature/src/feature/y.feature.ts
```diff
@@ -0,0 +1,11 @@
+import { COM, Environment, Feature, Service } from '@wixc3/engine-core';
+
+export const serverEnv = new Environment('server', 'node', 'single');
+
+export default new Feature({
+    id: 'YTestFeature',
```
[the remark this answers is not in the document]
olehrakwix: yes, @vladyslav-baliuk , please rename feature and all files to something like `default-args-echo`

packages/core/src/com/communication.ts
vladyslav-baliuk: @alisey fixed

packages/core/src/com/communication.ts
vladyslav-baliuk: @alisey fixed

packages/core/src/com/helpers.ts
vladyslav-baliuk: @alisey fixed

packages/core/src/com/helpers.ts
vladyslav-baliuk: @alisey fixed

test-fixtures/simple-feature/src/tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+      "outDir": "../dist"
+    },
+    "references": [
+      { "path": "../../../packages/core/src" },
+      {
+        "path": "../../../packages/core-node/src"
+      }
+    ]
+  }
+
```
olehrakwix, line 13: please add newline at the end of the file

test-fixtures/simple-feature/src/feature/y.feature.ts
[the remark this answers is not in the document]
vladyslav-baliuk: @alisey @olehrakwix it is just a simplified name for the feature that I'm using in test, actually the most of the features in `test-fixtures` have naming like that so I stayed consistent :)

packages/core/test/node/com.spec.ts
vladyslav-baliuk: @olehrakwix fixed

vladyslav-baliuk:
> > fix: added serialization/deserialization for arguments of the API calls
> 
> This is a not a reasonable title for this pull request.

fixed

test-fixtures/simple-feature/src/feature/y.feature.ts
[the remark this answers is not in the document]
olehrakwix: no, please still rename it so it has meaningful name

barak007:
this work is going to affect all messages in COM. do we really need to support this behavior? 
can't we try a typed approach so whenever we see a argument or return of `undefined` for an async service we use `undefined | null`

alisey:
> this work is going to affect all messages in COM. do we really need to support this behavior?

Yes, and what's the problem?

barak007:
It's kind of a hot spot. We need to keep an eye that we don't do unnecessary work.

alisey:
Vlad checked the impact on performance and didn't find any noticeable effect. (@vladyslav-baliuk , correct me if I'm wrong)

Additionally, we can do something like this:

```js
const args = [...];

function patchArguments(args) {
    for (let i = 0; i < args.length; i++) {
        if (args[i] === undefined) args[i] = undefinedPlaceholder;
    }
}
```

That way we wouldn't be creating extra work for the garbage collector. For most calls this would amount to merely looping through an array of 1-3 elements, which would amount to an overhead of ~0.00001 millisecond per remote API call.

But again, I'm not sure that would make any difference, modern generational garbage collector handle transient objects very efficiently.

test-fixtures/simple-feature/src/feature/y.feature.ts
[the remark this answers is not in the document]
vladyslav-baliuk: @olehrakwix fixed

vladyslav-baliuk:
> Vlad checked the impact on performance and didn't find any noticeable effect. (@vladyslav-baliuk , correct me if I'm wrong)
> 
> Additionally, we can do something like this:
> 
> ```js
> const args = [...];
> 
> function patchArguments(args) {
>     for (let i = 0; i < args.length; i++) {
>         if (args[i] === undefined) args[i] = undefinedPlaceholder;
>     }
> }
> ```
> 
> That way we wouldn't be creating extra work for the garbage collector. For most calls this would amount to merely looping through an array of 1-3 elements, which would add an overhead of ~0.00001 millisecond per remote API call. In other words, you would need to execute 100 000 remote API calls per second for this overhead to add up to a single millisecond.
> 
> But again, I'm not sure that would make any difference, modern generational garbage collector handle transient objects very efficiently.

Right, I've checked the amount of time needed for 

> Vlad checked the impact on performance and didn't find any noticeable effect. (@vladyslav-baliuk , correct me if I'm wrong)
> 
> Additionally, we can do something like this:
> 
> ```js
> const args = [...];
> 
> function patchArguments(args) {
>     for (let i = 0; i < args.length; i++) {
>         if (args[i] === undefined) args[i] = undefinedPlaceholder;
>     }
> }
> ```
> 
> That way we wouldn't be creating extra work for the garbage collector. For most calls this would amount to merely looping through an array of 1-3 elements, which would add an overhead of ~0.00001 millisecond per remote API call. In other words, you would need to execute 100 000 remote API calls per second for this overhead to add up to a single millisecond.
> 
> But again, I'm not sure that would make any difference, modern generational garbage collector handle transient objects very efficiently.

Correct, I've checked the amount of time needed to perform 100 remote API calls in a row, checked it on fresh machine several times per both of states - with and without serialization, no noticeable differences were found

Anyway I'll double-check it today with a bit different technique

packages/core/test/node/com.spec.ts
```diff
@@ -160,6 +160,25 @@ describe('Communication', () => {
         expect(handleMessageStub).to.have.not.been.called;
     });
 
+    it('serialization of arguments for remote API calls should handle undefined values ', async () => {
```
alisey, line 163: `supports default function parameters in remote API calls`

test-fixtures/default-args-echo/src/feature/feature.ts
```diff
@@ -0,0 +1,11 @@
+import { COM, Environment, Feature, Service } from '@wixc3/engine-core';
+
+export const serverEnv = new Environment('server', 'node', 'single');
+
+export default new Feature({
+    id: 'DefaultArgsEcho',
```
alisey, line 6: camel case

test-fixtures/default-args-echo/src/feature/server.env.ts
```diff
@@ -0,0 +1,9 @@
+import sampleFeature, { serverEnv } from './feature';
+
+sampleFeature.setup(serverEnv, () => {
+    return {
+        echoService: {
+            echo: (s = 'dude, it works!') => s,
```
alisey, line 6: Why do you need two echo services, `echo(s = 'No!')` and `echo(s = 'dude, it works!')`

packages/core/test/node/com.spec.ts
vladyslav-baliuk: @alisey fixed

test-fixtures/default-args-echo/src/feature/feature.ts
vladyslav-baliuk: @alisey fixed

test-fixtures/default-args-echo/src/feature/server.env.ts
vladyslav-baliuk: @alisey the first one is for testing serialization/deserialization itself but actually it should be without default value, just optional, fixed

The second one is a part of the feature created to test serialization in real remote API calls

alisey: I still don't understand why we need the first one if the second one is testing a real remote API call.

vladyslav-baliuk: @alisey the idea was to cover that functionality in two test-suits  - as part of `communication` tests to verify only serialization/deserialization process and as a part of `runtime-node` tests to verify how it works in real communication between envs

But you are right - the second test-case is pretty enough, fixed

test-fixtures/default-args-echo/src/feature/echo.server.env.ts
```diff
@@ -0,0 +1,9 @@
+import sampleFeature, { serverEnv } from './echo.feature';
```
alisey, line 1: Rename the import.

PR #1755, opened 2023-04-14
refactor: improve readability and fix electron-node.spec tests

olehrakwix:
improve naming as was requested here:
https://github.com/wixplosives/engine/pull/1718#pullrequestreview-1365311121

also, during working on this found that:
- handleUncaught option was not properly migrated to new implementation
- 'out-of-memory' error type is not used

so, added logic for handleUncaught option and removed 'out-of-memory' error type


alisey (approved):

test-fixtures/disconnecting-env/src/disconnecting-env.server.env.ts
```diff
@@ -1,14 +1,21 @@
 import testFeature, { serverEnv } from './disconnecting-env.feature';
 
-testFeature.setup(serverEnv, ({ run, errorType: { errorMode } }) => {
+testFeature.setup(serverEnv, ({ run, errorsConfig: { throwError, handleUncaught } }) => {
+    if (handleUncaught) {
+        process.on('uncaughtException', () => {
+            // User can do any other cleanup here
```
alisey, line 6: What does this comment mean? How can the user do cleanup in a test fixture?

olehrakwix: this was an old comment, I've just refactored this fixture. but I agree, this is useless, so removed it

PR #1769, opened 2023-04-24
refactor(electron-host): remove redundant listener

vladyslav-baliuk:
Remove `window-all-closed` listener that manually close the app when all windows are closed

### DIFF
--- a/packages/electron-host/src/start-electron-env.ts
+++ b/packages/electron-host/src/start-electron-env.ts
@@ -1,5 +1,5 @@
 import fs from '@file-services/node';
-import { app, BrowserWindow, ipcMain } from 'electron';
+import { app, ipcMain } from 'electron';
 
 import { BaseHost, Environment, RuntimeEngine, TopLevelConfig } from '@wixc3/engine-core';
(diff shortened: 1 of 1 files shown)


packages/electron-host/src/start-electron-env.ts
```diff
@@ -107,22 +107,5 @@ export async function runElectronEnv({
...
-
-        if (BrowserWindow.getAllWindows().length === 0) {
-            try {
-                await runningEnvironment.dispose();
-            } catch (e) {
-                process.exitCode = 1;
-                // eslint-disable-next-line no-console
-                console.error(e);
-            }
-            app.quit();
```
olehrakwix, line 123: how does quite event triggered now?

vladyslav-baliuk: @olehrakwix if `window-all-closed` listener has no subscribers the default behavior would be to quit the app, thus trigger the event (according to [doc](https://www.electronjs.org/docs/latest/api/app#event-window-all-closed))

alisey (approved):

PR #1782, opened 2023-05-02
fix: workerthreads support in testkit

olehrakwix:


packages/electron-commons/test/initializers/electron-node.spec.ts
```diff
@@ -35,9 +35,9 @@ describe('onDisconnectHandler for node environment initializer', () => {
             });
 
             await dispose();
-
-            await expect(exitPromise).to.eventually.deep.eq({ exitCode: 0 });
+            await expect(exitPromise).to.eventually.be.fulfilled;
```
AviVahl, line 38: previous is better. why switch?

packages/runtime-node/src/worker-thread-entry.ts
```diff
@@ -53,7 +53,11 @@ const handleStartupMessage = async (command: WorkerThreadStartupCommand) => {
         childEnvName: environmentContextName,
     });
 
-    disposeNodeEnv = runningNodeEnv.dispose;
+    disposeNodeEnv = () => {
+        ensureWorkerThreadContext(parentPort);
```
AviVahl, line 57: this was already validated during start. use `!` in next line instead?

packages/runtime-node/src/worker-thread-entry.ts
```diff
@@ -65,7 +69,7 @@ const messageHandler = (message: unknown) => {
             break;
 
         case 'workerThreadDisposeCommand':
-            Promise.all([disposeNodeEnv])
+            Promise.all([disposeNodeEnv()])
```
AviVahl, line 72: is the `Promise.all` even needed?

packages/test-kit/test/run-environment.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+        const { dispose, runningApi } = await getRunningFeature({
+            env: serverEnv,
+            feature: workerThreadFeature,
+            featureName: workerThreadFeature.id,
+            basePath: featurePath,
+        });
+
+        const result = await runningApi.workerService.initAndCallWorkerEcho('hello');
+        expect(result).to.equal('hello from worker');
+        await dispose();
```
AviVahl, line 21: if expect above fails, this won't be called. use createDisposables

packages/test-kit/package.json
```diff
@@ -4,7 +4,9 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "test": "mocha \"test/**/*.unit.ts?(x)\""
+    "test": "yarn test:unit && yarn test:spec",
```
AviVahl, line 7: why is the separation needed? call the test file `.unit`

AviVahl (changes requested):

packages/electron-commons/test/initializers/electron-node.spec.ts
olehrakwix: it gets flaky for some reasons. somehow on CI it sometimes has { signal: SIGTERM }, and others { exitCode: 0 }
does not reproduce locally

packages/runtime-node/src/worker-thread-entry.ts
olehrakwix: fixed

packages/test-kit/test/run-environment.spec.ts
olehrakwix: done

packages/test-kit/package.json
olehrakwix: ok, renamed

AviVahl (approved):

PR #1793, opened 2023-05-05
fix!: make worker initializer sync and deffered

olehrakwix:
- make workerThreadInitializer sync to immediately return dispose function
- fix start in server mode to provide metadata
- update metadata provider to be object and have dispose method


packages/core-node/src/create-application-metadata-provider.ts
```diff
@@ -0,0 +1,69 @@
...
+import { METADATA_PROVIDER_ENV_ID, ENGINE_ROOT_ENVIRONMENT_ID } from './constants';
+import memoizeOne from 'memoize-one';
+
+/**
+ * creates a new instance of metadata provider that can get application metadata using `MetadataCollectionAPI` api
+ * that should be registered for `ROOT_ENGINE_ENV_ID` environment
+ * @param com The communication instance that have registered hosts for `ROOT_ENGINE_ENV_ID` and `METADATA_PROVIDER_ENV_ID`
+ */
+export function createMetadataProvider(com: Communication) {
+    const metadataProviderCom = createMetadataProviderCommunication(com);
```
cylon3035, line 13: can you use `createMetadataProviderCommunication` inside `getMetadataFromCommunication` si we do not need to memoize twice?

olehrakwix: moved all logic to single function

cylon3035 (approved):

PR #1808, opened 2023-05-14
refactor: remove application proxy (merge packages pre-step)

barak007:


barak007:
mostly not needed in the new flow

PR #1814, opened 2023-05-16
allow injecting topology using a global variable, $CODUX_TOPOLOGY

nadav-ab:
allow injecting topology to codux entry point using a global variable "$CODUX_TOPOLOGY"
if it exists it replaces the data collected from the url search params

this will allow us to inject the topology by making sure a previous script has set it

### DIFF
--- a/packages/scripts/src/create-entrypoint.ts
+++ b/packages/scripts/src/create-entrypoint.ts
@@ -136,21 +136,26 @@ async function main() {
     const topWindow = currentWindow.parent ?? currentWindow;
     const isMainEntrypoint = topWindow && currentWindow === topWindow;
     const options = new URLSearchParams(currentWindow.location.search);
+    const optionsObj = typeof $CODUX_TOPOLOGY !== undefined ? $CODUX_TOPOLOGY : {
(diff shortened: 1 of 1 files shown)


barak007 (changes requested):
No codux here!

packages/scripts/src/create-entrypoint.ts
```diff
@@ -136,21 +136,26 @@ async function main() {
     const topWindow = currentWindow.parent ?? currentWindow;
     const isMainEntrypoint = topWindow && currentWindow === topWindow;
     const options = new URLSearchParams(currentWindow.location.search);
+    const optionsObj = typeof $CODUX_TOPOLOGY !== undefined ? $CODUX_TOPOLOGY : {
+        publicPath: options.get('publicPath'),
+        ${FEATURE_QUERY_PARAM}: options.get('${FEATURE_QUERY_PARAM}'),
```
barak007, line 141: this can generate invalid code. let's talk about another way to inject it. maybe via `globalThis.entryOptions`

barak007:
#1818

PR #1819, opened 2023-05-17
fix(test-kit): grant `clipboard-read` permissions for `playwright`

vladyslav-baliuk:
Needed in order to allow `playwright` retrieve the text stored in clipboard when running a feature in test mode


olehrakwix (approved):

AviVahl:
We already tried it in the past and it does not work on all platforms, if I remember correctly...

PR #1828, opened 2023-05-24
fix: move identify apis to feature construction

barak007:
We have tests that are reling on the identification to happen before running the feature.

This can be a pitfall if someone would share these entities between features. anyhow this is not new so only fixing the issue.


barak007:
this will not work

PR #1838, opened 2023-05-29
fix: add safari polyfill for nested web workers

olehrakwix:


packages/runtime-node/src/worker-thread-entry.ts
```diff
@@ -55,13 +54,13 @@ const handleStartupMessage = async (command: WorkerThreadStartupCommand) => {
...
     disposables.add(() => {
-        parentPort!.off('message', messageHandler);
+        worker.removeEventListener('message', messageHandler);
         return engine.shutdown();
     });
 };
 
-const messageHandler = (message: unknown) => {
-    const workerThreadCommand = message as WorkerThreadCommand;
+const messageHandler = (message: any) => {
```
barak007, line 62: why this changed to any?

.github/workflows/tests.yml
```diff
@@ -15,8 +15,11 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
+          registry-url: 'https://registry.npmjs.org/'
```
barak007, line 18: why do we need this?

barak007 (commented):
Looks good. few I left couple of questions.

.github/workflows/tests.yml
olehrakwix: Cause @wixc3/isomorphic-worker lives there, and I got these errors without this change:
https://github.com/wixplosives/engine/actions/runs/5110105664/jobs/9185624849#step:6:13

packages/runtime-node/src/worker-thread-entry.ts
olehrakwix: changed back to `unknown`

packages/core/src/com/hosts/universal-worker-host.ts
```diff
@@ -0,0 +1,32 @@
+import type {
+    UniversalMessageHandler,
+    UniversalWorker,
+    UniversalWorkerUserMethods,
+} from '@wixc3/isomorphic-worker/dist/types';
```
AviVahl, line 5: avoid importing directly from dist

AviVahl (changes requested):
the approach here uses a private package

packages/core/src/com/hosts/universal-worker-host.ts
olehrakwix: fixed

AviVahl (approved):

PR #1868, opened 2023-06-15
fix(com): use unique message ids

cylon3035:
Communication forward message functionality can accidentally resolve wrong pending callback record if callbackId is the same. Since previously call and callbacks ids were just counters `c0` `c1` `c2` - the chances of it happening are high at the startup.

p.s. forward message functionality seemed wrong but math checks out. Just callback ids should be more unique. Environment id is nor improving uniqueness, but come in handy for debugging. 

fixes issue with workers not starting sometimes

### DIFF
--- a/packages/core/src/com/communication.ts
+++ b/packages/core/src/com/communication.ts
@@ -98,7 +98,7 @@ export class Communication {
         this.registerMessageHandler(host);
         this.registerEnv(id, host);
         this.environments['*'] = { id, host };
-        this.messageIdPrefix = `c_${this.getEnvironmentId()}_${Math.random().toString(36)}`;
+        this.messageIdPrefix = `c_${this.getEnvironmentId()}_${Math.random().toString(36).slice(2)}`;
         this.post(this.getPostEndpoint(host), {
             type: 'ready',
             from: id,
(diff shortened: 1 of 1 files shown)


AviVahl (commented):
Math.random() is a bit too long, isn't it?

vlad-karnauch:
@AviVahl How many characters would you like?

AviVahl:
6-8 alpha numeric? Remember, this is added to every message and will increase overhead.

packages/core/src/com/communication.ts
```diff
@@ -98,7 +98,7 @@ export class Communication {
         this.registerMessageHandler(host);
         this.registerEnv(id, host);
         this.environments['*'] = { id, host };
-
+        this.messageIdPrefix = `c_${this.getEnvironmentId()}_${Math.random()}`;
```
barak007, line 101: too long. the environment id is already on the message. also where did you encounter an issue with the callback ids?

barak007: if the issue is for environment in the same runtime...you could add this to the constructor and use it as prefix. this will produce a small callback id.

```ts
Communication.instanceCount ||=0
Communication.instanceCount++;
```

AviVahl (approved):

PR #1896, opened 2023-06-27
test: parallel run from root; ix -> it

AviVahl:
and a bit more consistent "test" script names


barak007 (approved):

AviVahl:
there was flakiness in the workerthread test when running in parallel in CI

barak007:
I think I know what's the issue. It does not dispose the env. I add this dispose to my initializer today and noticed that. Maybe that's the issue

PR #1907, opened 2023-07-04
refactor: reintroduce the new entry point

barak007:


packages/core/src/runtime-main.ts
```diff
@@ -0,0 +1,61 @@
...
+        (globalThis as any).name = instanceId;
+    }
+
+    const buildConfigPromise = runtimeConfiguration.importConfig(configName);
+    const runtimeConfigPromise = runtimeConfiguration.load(env.env, featureName, configName);
+
+    const [buildConfig, runtimeConfig] = await Promise.all([buildConfigPromise, runtimeConfigPromise]);
+
+    // only load features after the config is loaded to avoid blocking onload event
+    const { entryFeature, resolvedContexts } = await featureLoader.loadEntryFeature(featureName, {});
```
barak007, line 51: this was the pain point

packages/runtime-node/src/node-environment.ts
```diff
@@ -47,22 +47,9 @@ export async function runNodeEnvironment<ENV extends AnyEnvironment>({
...
-    const featureLoader = new FeatureLoadersRegistry(new Map(Object.entries(featureLoaders)), resolvedContexts);
-    const optionsRecord: Record<string, string | boolean> = {};
 
-    for (const [key, val] of options || []) {
-        optionsRecord[key] = val;
-    }
-    const loadedFeatures = await featureLoader.getLoadedFeatures(featureName, optionsRecord);
-    const runningFeatures = [loadedFeatures[loadedFeatures.length - 1]!];
+    const featureLoader = new FeatureLoadersRegistry(new Map(Object.entries(featureLoaders)));
+    const { entryFeature, resolvedContexts } = await featureLoader.loadEntryFeature(featureName, toRecord(options));
```
daomry, line 52: why not use `Object.fromEntries(options || [])` ?

packages/runtime-node/src/node-environment.ts
```diff
@@ -138,3 +125,11 @@ export function createFeatureLoaders(
     }
     return featureLoaders;
 }
+
+function toRecord(options: [string, string | boolean][] | undefined) {
```
daomry, line 129: very close to a Object.fromEntries clone, can it be removed?

packages/core/src/run-engine-app.ts
```diff
@@ -78,7 +75,8 @@ export class FeatureLoadersRegistry {
      */
     async getLoadedFeatures(
         rootFeatureName: string,
-        runtimeOptions: Record<string, string | boolean> = {}
+        runtimeOptions: Record<string, string | boolean>,
+        resolvedContexts: Record<string, string>
```
daomry, line 79: now that the change is tested, can we keep the old defaults API (rootFeatureName , ...={}, ...={}) ?

packages/core/src/runtime-configurations.ts
```diff
@@ -0,0 +1,138 @@
...
+
+export type ConfigModule = {
+    default: TopLevelConfig;
+};
+
+export type ConfigLoader = () => Promise<ConfigModule[]>;
+
+export type ConfigLoaders = Record<string, ConfigLoader>;
+
+export class RuntimeConfigurations {
```
daomry, line 17: plz add comment explain what this class is used for an how... and read what gpt writes

daomry (approved):

daomry (changes requested):
plz add ts-doc comments for exported classes/functions

packages/runtime-node/src/node-environment.ts
barak007: done

packages/runtime-node/src/node-environment.ts
barak007: removed

packages/core/src/runtime-configurations.ts
barak007: added comments to each public function and the class itself

daomry (approved):

PR #1940, opened 2023-07-25
feat: persist page option to with-feature 

matangeorgi:
Adding to the with-feature the option to keep the page open through all the tests in the suite.

Example of the persist option usage:
```ts
describe('Persist option', function () {
    this.timeout(20_000);

    const { getLoadedFeature } = withFeature({
        featureName: 'react/someplugin',
        runOptions: {
            projectPath: join(__dirname, '..'),
        },
        persist: true,
    });

    let page: Page;
    before(async function () {
        ({ page } = await getLoadedFeature());
    });

    it('...', async () => {
 // No need to call getLoadedFeature again to initialize the page.
    });
});
```


packages/test-kit/src/with-feature.ts
```diff
@@ -255,43 +260,70 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
...
+                runningFeature = await executableApp.runFeature({
+                    featureName,
+                    configName,
+                    runtimeOptions: runOptions,
+                    overrideConfig: config,
+                });
+            };
+
+            if (persist) {
+                before('load feature', runFeature);
```
AviVahl, line 274: we're inside getLoadedFeature, which means we're inside an it, so we shouldn't call `before` here. hooking into before should be in withFeature closure (which is executed in describe).

packages/test-kit/src/with-feature.ts
```diff
@@ -255,43 +260,70 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}) {
                 throw new Error('Engine HTTP server is closed!');
             }
 
-            const runningFeature = await executableApp.runFeature({
-                featureName,
-                configName,
-                runtimeOptions: runOptions,
-                overrideConfig: config,
-            });
+            let runningFeature;
```
AviVahl, line 263: type?

packages/test-kit/src/with-feature.ts
```diff
@@ -109,6 +109,10 @@ export interface IWithFeatureOptions extends Omit<IFeatureExecutionOptions, 'tra
      * add tracing for the entire suite, the name of the test will be used as the zip name
      */
     tracing?: boolean | Omit<Tracing, 'name'>;
+    /**
+     * Keeps the page open for the all the tests in the suite
```
AviVahl, line 113: not just the page, the feature, right?

packages/test-kit/src/with-feature.ts
AviVahl: also, this pattern of overriding a local variable leaves a lot of room for errors. use an array/set of running features

AviVahl (changes requested):
the location of the `before` calls is incorrect. revise code.

packages/test-kit/src/with-local-fixture.ts
```diff
@@ -22,7 +23,15 @@ export function withLocalFixture(suiteOptions: IWithLocalFixtureOptions) {
...
-        const projectPath = createTestDir('local-test');
+        // const projectPath = createTestDir('local-test');
+        const dir = createTempDirectorySync('local-test');
+        const { persist } = testOptions;
+        if (persist) {
+            after(() => dir.remove());
+        } else {
+            disposeAfter(() => dir.remove());
+        }
+        const projectPath = fs.realpathSync.native(dir.path);
```
AviVahl, line 34: no need for the realpathSync anymore.

AviVahl (approved):

PR #2012, opened 2023-08-20
docs: add docusaurus to existing *.md docs

plisovyi:
I've restructured the existing docs in `/docs`, added docusaurus 2, and set up build and deployment alongside existing API reference docs.

`sudo` is used in commands because GitHub says it's ok to use, and this is the only repo we have that uses actions for deployment to GitHub pages rather than deployment from a branch.

Some documentation may still be outdated, I've mostly prepared it for deployment and plan to update it with the deployment in place.

How it will look after merge/deployment could be seen at https://plisovyi.github.io/engine/docs/


AviVahl:
there should not be a separate yarn.lock for docs

AviVahl:
also, docusaurus has (never-fixed) security issues AFAIK. this repo currently has:
```
0 vulnerabilities found - Packages audited: 1327
Done in 0.98s.
```

A fresh docusaurus installation has:
```
43 vulnerabilities (13 moderate, 26 high, 4 critical)
```

plisovyi:
> ```
> 22 vulnerabilities (9 moderate, 13 high)
> ```

Should we just have it as a separate repository then?

PR #2013, opened 2023-08-21
fix: add hostname

oleksandraab:
fix for https://github.com/wixplosives/codux/issues/13458


packages/runtime-node/src/launch-http-server.ts
```diff
@@ -21,13 +21,15 @@ export interface ILaunchHttpServerOptions {
     httpServerPort?: number;
     socketServerOptions?: Partial<io.ServerOptions>;
     routeMiddlewares?: Array<RouteMiddleware>;
+    hostname?: string | undefined;
```
cylon3035, line 24: no need to do  `| undefined` when we have `?`

cylon3035 (changes requested):

cylon3035 (approved):

PR #2035, opened 2023-09-12
feat: avoid analyzing type only imports and exports

barak007:
This will improve feature analyze pref

### DIFF
--- a/packages/scripts/src/utils/resolve-module-graph.ts
+++ b/packages/scripts/src/utils/resolve-module-graph.ts
@@ -60,8 +60,8 @@ function extractModuleRequests(sourceFile: ts.SourceFile): string[] {
     };
 
     const importsFinder = (node: ts.Node) => {
-        const isNonTypeImport = isImportDeclaration(node) && !node.importClause?.isTypeOnly;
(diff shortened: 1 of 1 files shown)


packages/scripts/src/utils/resolve-module-graph.ts
```diff
@@ -78,6 +78,26 @@ function extractModuleRequests(sourceFile: ts.SourceFile): string[] {
...
+    return node.isTypeOnly || hasOnlyTypeBindings(node.exportClause);
+}
+
+function isTypeOnlyImport(node: ts.ImportDeclaration) {
+    return node.importClause?.isTypeOnly || hasOnlyTypeBindings(node.importClause?.namedBindings);
+}
+
+function hasOnlyTypeBindings(bindings?: ts.NamedImportBindings | ts.NamedExportBindings) {
+    if (bindings?.kind === ts.SyntaxKind.NamedImports || bindings?.kind === ts.SyntaxKind.NamedExports) {
+        for (const element of bindings.elements) {
```
AviVahl, line 91: `return bindings.elements.every(el => el.isTypeOnly)`

AviVahl: if u want perf, extract the arrow fn to top level

PR #2176, opened 2023-11-30
fix: remove usage of global disposeAfter from withFeature

barak007:


packages/test-kit/src/with-feature.ts
```diff
@@ -561,26 +559,22 @@ async function enableTracing({
...
-                fs,
-                name:
-                    process.env.TRACING && process.env.TRACING !== 'true'
-                        ? process.env.TRACING
-                        : name ?? (testName ? normalizeTestName(testName) : 'nameless-test'),
-            }),
-        });
-    });
     dispose(
         async () => {
```
matangeorgi, line 563: the `async` is needed here?

matangeorgi:
good change, organized a lot of stuff.

packages/test-kit/src/with-feature.ts
barak007: not needed

packages/test-kit/src/run-environment.ts
```diff
@@ -1,25 +1,24 @@
...
     /**@deprecated use engine.shutdown */
     dispose: () => Promise<void>;
 }> {
     const runningFeature = await originalGetRunningFeature(options);
-    if (disposeAfterTestTimeout) {
-        disposeAfter(runningFeature.engine.shutdown, {
-            name: `engine shutdown for ${options.featureName}`,
-            timeout: disposeAfterTestTimeout,
+    if (autoDisposeTimeout) {
+        afterEach(`engine shutdown for ${options.featureName}`, function () {
```
barak007, line 19: protect `afterEach`

PR #2178, opened 2023-12-01
fix(withFeature): disable-failed-tests-screenshots-by-default

zetmate:
Added an env variable to enable/disable failed tests screenshots, also made disabled by default.
Related [this](https://github.com/orgs/wixplosives/projects/27/views/19?pane=issue&itemId=46322716)


zetmate:
> Why we want to disable this by default?
@barak007 I've heard from @plisovyi that locally it's very annoying that they get generated constantly every time you run the tests that fail.

zetmate:
@barak007 actually it seems like I forgot to put it into gitignore... at least it's not in the codux nor engine

packages/test-kit/src/with-feature.ts
```diff
@@ -225,7 +227,7 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}): WithF
         devtools = envDebugMode ? debugMode : undefined,
         slowMo,
         persist,
-        takeScreenshotsOfFailed = true,
+        takeScreenshotsOfFailed = process.env.TESTS_SCREENSHOTS === 'true',
```
Ari-Kishon, line 230: Maybe "TEST_SCREENSHOTS"

Also lets add some documentation for this environment variable

plisovyi:
If it's in .gitignore, I guess it will be fine, but it makes it required on the consumer's repo side for such a default 🤷🏼

barak007:
we already have few directories that should be ignored that the engine generates. we could make it more verbose but with proper docs it is not an issue.
I don't mind the env variable I just question the default. 
it's my opinion if someone has other strong opinions we can talk it out.

plisovyi:
I'm personally okay with just updating codux .gitignore, and updating docs in the engine about it.

barak007:
Please open new one if you still want the env variable. I think we don't change the default flag

PR #2192, opened 2023-12-07
fix: added reportError polyfil

olehrakwix:
Added reportError polyfill to allow applications and are using native `reportError` to execute feature setup.

### DIFF
--- a/packages/core/src/polyfills/report-error-polyfill.ts
+++ b/packages/core/src/polyfills/report-error-polyfill.ts
@@ -1,9 +1,12 @@
 import { reportError } from "../com";
 
 /**
- * hope this is single polyfill we need for now
- * in case if more will be added, we need to create some
- * better mechanism to handle polyfills
+ * This polyfill is not essential for the engine itself. But applications built
+ * on the engine may require this polyfill to correctly report initialization
+ * errors in older browsers.
+ * Ideally, the engine should offer a mechanism allowing applications to
+ * initialize necessary polyfills before the evaluation of feature files begins.
+ * But so far we only need this single polyfill.
  */
 if (globalThis.reportError === undefined) {
     globalThis.reportError = reportError;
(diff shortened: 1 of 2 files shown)


packages/core/src/runtime-main.ts
```diff
@@ -1,3 +1,5 @@
+import "./polyfills/report-error-polyfill.js";
```
alisey, line 1: No need for it in web workers?

packages/core/src/polyfills/report-error-polyfill.ts
```diff
@@ -0,0 +1,10 @@
+import { reportError } from "../com";
+
+/**
+ * hope this is single polyfill we need for now
+ * in case if more will be added, we need to create some
+ * better mechanism to handle polyfills
+ */
```
alisey, line 7:
```ts
/**
 * This polyfill is not essential for the engine itself. But applications built
 * on the engine may require this polyfill to correctly report initialization
 * errors in older browsers.
 * Ideally, the engine should offer a mechanism allowing applications to
 * initialize necessary polyfills before the evaluation of feature files begins.
 * But so far we only need this single polyfill.
 */
```

alisey (approved):

packages/core/src/runtime-main.ts
olehrakwix: no, I verified that this entrypoint is run for any type of environment, including webworker

packages/core/src/polyfills/report-error-polyfill.ts
olehrakwix: thanks, updated

AviVahl:
I'm removing this, as this stops the engine-core package from having no side effects.

We need to find a better way to inject polyfills needed by applications.
It should not come as a hack within engine-core, which also caused this method to be put in Node environments.

As a replacement to this change, I'm changing engine-core's `reportError` to default to `globalThis.reportError`, with the current implementation used as a fallback. This can be imported and used.

PR #2238, opened 2023-12-31
feat: support runtimeArgs from cli

barak007:
allow passing runtime args as json to the new esbuild cli


idoros (approved):
I don't like passing json as argument, but I assume this will be changed once we build flows for developer that actually uses the new flow

PR #2239, opened 2023-12-31
fix: run tests in fresh context by default

barak007:
This PR adds a `resetContextBetweenTests` configuration flag for `withFeature` to set if the browser context is recreated for each test or shared between tests - defaults to `true`


packages/test-kit/src/with-feature.ts
```diff
@@ -140,6 +140,10 @@ export interface IWithFeatureOptions extends Omit<IFeatureExecutionOptions, 'tra
      * Prebuild the engine before running the tests
      */
     buildFlow?: 'prebuild' | 'lazy';
+    /**
+     * fresh context for each test
+     */
+    freshContext?: boolean;
```
idoros, line 146: createContextPerTest maybe?

idoros (approved):

PR #2243, opened 2024-01-02
refactor: try to simplify and remove module state from `createMetadataProvider`

barak007:
This PR simplify the metadata fetching for for `renderer` and `workerthread` initializers


barak007:
this is going to be dead code.

PR #2251, opened 2024-01-10
IDisposable + SafeDisposable refactor

daomry:
- Using IDisposable from wixc3/patterns 
- removed engine IDisposable
- Using SafeDispose for disposables


packages/core/src/runtime-feature.ts
```diff
@@ -7,41 +7,38 @@ import {
...
     private disposeHandlers = new SetMultiMap<string, DisposeFunction>();
-    private disposing: IDeferredPromise<void> | undefined;
     constructor(
         public feature: InstanceType<T>,
         public api: Running<T, ENV>,
         public dependencies: RunningFeatures<InstanceType<T>['dependencies'], ENV>,
         public environment: ENV,
-    ) {}
+    ) {
+        this.disposables.add('disposal handlers', async () => {
```
barak007, line 29: does this cover the this.disposing?

packages/runtime-node/package.json
```diff
@@ -29,4 +29,4 @@
   "publishConfig": {
     "access": "public"
   }
-}
+}
```
barak007, line 32: how do you change these files? do you run `pleb upgrade?`

packages/runtime-node/src/core-node/ws-node-host.ts
```diff
@@ -13,16 +14,17 @@ export class WsHost extends BaseHost {
...
     private socketToEnvId = new Map<string, { socket: io.Socket; clientID: string }>();
-    private disposed = false;
+    private disposables = new SafeDisposable(WsServerHost.name);
+    dispose = this.disposables.dispose;
+    isDisposed = this.disposables.isDisposed;
 
     constructor(private server: io.Server | io.Namespace) {
         super();
         this.server.on('connection', this.onConnection);
+        this.disposables.add('connection', () => this.server.off('connection', this.onConnection));
```
barak007, line 26: why two hops? it's just removing listeners. I feel that the pattern should NOT be add disposable for each clear. WDYF?

packages/runtime-node/src/node-env-manager.ts
```diff
@@ -78,11 +84,16 @@ export class NodeEnvManager {
 
         const host = new WsServerHost(socketServer);
 
-        this.autoLaunchDispose = async () => {
-            host.dispose();
-            disposeListener();
-            await close();
-        };
+        this.disposables.registerGroup('auto launch', { before: 'default' });
```
barak007, line 87: why group here? it will dispose after every time because we register in constructor.

packages/runtime-node/src/core-node/ipc-host.ts
```diff
@@ -24,6 +27,8 @@ export class IPCHost extends BaseHost implements IDisposable {
                 });
             }
         });
+        this.disposables.add('process listeners', () => this.process.removeAllListeners());
```
barak007, line 30: why two separated adds?

barak007 (commented):
The changes themself looks fine in terms of "will it work", but in general it feels like overkill pattern in some places.

packages/core/src/runtime-feature.ts
daomry: yes, built in to SafeDisposable

packages/runtime-node/package.json
daomry: not in this chase, since I wanted to only update core3utils libs

barak007 (approved):

PR #2252, opened 2024-01-10
deps upgrade - core3utils

daomry:


barak007 (approved):

AviVahl:
I've upgraded `master` with these ones and the rest as well.
Start using prettier please, and make sure your IDE uses it for .json files. EOL getting removed from every package.json in this PR is an indication that another formatter is being used.

PR #2270, opened 2024-01-15
fix: try fix performance issue when running in tests

barak007:
Try to fix performance issues for tests that uses `getRunningFeature`

### DIFF
--- a/packages/scripts/src/run-environment.ts
+++ b/packages/scripts/src/run-environment.ts
@@ -194,6 +194,9 @@ function locateEnvironment(
     return undefined;
 }
 
+/**
+ * this is used for running tests in node and only for tests
+ */
 export async function getRunningFeature<F extends FeatureClass, ENV extends AnyEnvironment>(
     options: RunningFeatureOptions<F, ENV>,
 ): Promise<{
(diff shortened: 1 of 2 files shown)


packages/scripts/src/run-environment.ts
```diff
@@ -46,17 +46,18 @@ export interface RunningFeatureOptions<F extends FeatureClass, ENV extends AnyEn
...
-    bundlePath,
-    runtimeOptions = {},
-    config = [],
-    env,
-    basePath = process.cwd(),
-    featureDiscoveryRoot,
-}: IRunNodeEnvironmentOptions<ENV>): Promise<RuntimeEngine<ENV>> {
-    const { env: envName, envType } = env;
+function memoize<T extends (...args: string[]) => any>(fn: T): T {
+    const cache = new Map<string, ReturnType<T>>();
```
cylon3035, line 50: WeakMap ?

barak007: not for strings...

barak007: also there is not reason here it will only run once and during tests.

cylon3035 (approved):

alissawix (approved):

PR #2279, opened 2024-01-18
fix(core): don't allow native Object methods to be proxied

barak007:
When debugging the debugger calls `toString` and `valueOf` functions on the proxies. that causes messages to fly around when not expected.

This PR fixes that by not proxy built-in `Object.prototype` methods 
I also added a guard to the `AsyncApi` type to make these functions `never`


idoros (approved):
I think it would be better not to fall back on the object defaults, and instead return a value that reflects that it isn't the actual result, at least for `toString` it could have been something like `toString is not supported as a function between environments`

PR #2280, opened 2024-01-18
feat: added specific disconnect error

barak007:
Since most places want to ignore a disconnect error now they can do it easily

* Removed unnecessary `handleReady` call when calling `envReady` add an environment to pending state
* Better error when rejecting a callback (include origin message)

### DIFF
--- a/packages/core/src/com/communication.ts
+++ b/packages/core/src/com/communication.ts
@@ -938,7 +938,7 @@ export class Communication {
                 this.pendingCallbacks.delete(callbackId);
                 delete this.callbacks[callbackId];
                 clearTimeout(timerId);
-                error.stack += `\nCaused by: ${JSON.stringify(cleanMessageForLog(message), null, 2)}`;
+                error.message += `\nCaused by: ${JSON.stringify(cleanMessageForLog(message), null, 2)}`;
                 rej(error);
             };
             if (this.options.warnOnSlow) {
--- a/packages/core/src/com/communication.ts
+++ b/packages/core/src/com/communication.ts
@@ -55,6 +55,7 @@ import {
     type Target,
     type UnknownFunction,
 } from './types.js';
+import { EnvironmentDisconnectedError } from './environment-disconnected-error.js';
 
 export interface ConfigEnvironmentRecord extends EnvironmentRecord {
     registerMessageHandler?: boolean;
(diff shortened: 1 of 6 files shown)


packages/core/src/com/environment-disconnected-error.ts
```diff
@@ -0,0 +1,6 @@
+export class EnvironmentDisconnectedError extends Error {
+    constructor(message: string) {
+        super(message);
+        this.name = 'EnvironmentDisconnectedError';
```
AviVahl, line 4: use new.target? https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#support-for-newtarget

packages/core/src/com/communication.ts
```diff
@@ -941,6 +938,7 @@ export class Communication {
                 this.pendingCallbacks.delete(callbackId);
                 delete this.callbacks[callbackId];
                 clearTimeout(timerId);
+                error.stack += `\nCaused by: ${JSON.stringify(cleanMessageForLog(message), null, 2)}`;
```
AviVahl, line 941: wouldn't this affect source maps support? why add it to the stack (and trigger stack generation) and not to the .message?

packages/core/src/com/environment-disconnected-error.ts
barak007: I don't need this here. I don't have any extra prototype chain

packages/core/src/com/communication.ts
barak007: It can be in the message. trigger stack generation ? what do you mean? source map support still works.

idoros (approved):

packages/core/src/com/communication.ts
```diff
@@ -941,6 +938,7 @@ export class Communication {
                 this.pendingCallbacks.delete(callbackId);
                 delete this.callbacks[callbackId];
                 clearTimeout(timerId);
+                error.message += `\nCaused by: ${JSON.stringify(cleanMessageForLog(message), null, 2)}`;
```
alisey, line 941: @barak007 @idoros 

These are error messages from cross-environment calls, they're shown to the users, and now they look like this:

image



I'm removing this, it's not acceptable. There's a comment in this very file about the importance of not tampering error messages:
https://github.com/wixplosives/engine/blob/80d420d3b91d0170650b090adf4d6c3203094a14/packages/core/src/com/communication.ts#L869-L871

PR #2291, opened 2024-01-29
feat: engine dashboard

barak007:
This PR implements a new simple dashboard for the engine


packages/engine-cli/src/engine-build.ts
```diff
@@ -176,17 +177,33 @@ export async function runEngine({
             console.log(`Runtime options: ${JSON.stringify(Object.fromEntries(runtimeOptions.entries()), null, 2)}`);
         }
 
-        runManagerContext = runNodeManager({
```
idoros, line 179: runNodeManager doesn't seems to be used anymore

packages/engine-cli/src/dashboard/dashboard.ts
```diff
@@ -0,0 +1,119 @@
...
+        tr.appendChild(configName);
+        tr.appendChild(run);
+
+        featureTable.appendChild(tr);
+    });
+}
+
+function populateDatalistOptions(options: string[], id: string) {
+    const dataList = byId(id);
+    dataList.id = id;
```
idoros, line 98: getting by id and then re-setting it?

packages/engine-cli/src/engine-build.ts
barak007: removed

packages/engine-cli/src/dashboard/dashboard.ts
barak007: removed

idoros (approved):

PR #2299, opened 2024-02-01
fix: few issues with runtime args and restart

barak007:
* validate runtimeArgs before doing anything
* restart wait for dispose of the previous to get the "same" port.
* better errors 
* sticky headers

### DIFF
--- a/packages/engine-cli/src/engine-build.ts
+++ b/packages/engine-cli/src/engine-build.ts
@@ -309,7 +309,7 @@ async function launchDevServer(
     );
     const autoRunFeatureName = runtimeOptions.get('feature') as string | undefined;
     if (autoRunFeatureName) {
-        const port = await run(autoRunFeatureName, runtimeOptions.get('config') as string, '');
+        const port = await run(autoRunFeatureName, runtimeOptions.get('config') as string, '', false);
         // TODO: get the names of main entry points from the build configurations
         console.log(`Engine application in running at http://localhost:${port}/main.html`);
     } else {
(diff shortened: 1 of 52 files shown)


packages/engine-cli/src/engine-build.ts
```diff
@@ -377,10 +383,15 @@ function runOnDemandSingleEnvironment(
...
-        );
-        run(req.body.featureName, req.body.configName, req.body.runtimeArgs)
+        let message = `running on demand feature: "${req.body.featureName}" config: "${req.body.configName}"`;
+        if (req.body.runtimeArgs) {
+            message += ` runtimeArgs: "${req.body.runtimeArgs}"`;
+        }
+        if (req.body.restart) {
+            message += ' with restart';
+        }
+        console.log(message);
```
idoros, line 393: remove log

packages/engine-cli/src/engine-build.ts
```diff
@@ -377,10 +383,15 @@ function runOnDemandSingleEnvironment(
...
+        run(req.body.featureName, req.body.configName, req.body.runtimeArgs, req.body.restart)
```
idoros, line 394: restart can be undefined

packages/engine-cli/src/engine-build.ts
barak007: The log is the point here. It should be there

packages/engine-cli/src/engine-build.ts
barak007: Yep will fix

barak007: It does not really matter because we only care about the truth case. I will consider removing it and always wait

PR #2305, opened 2024-02-04
feat: better dev

barak007:
* !change tests to run in esbuild prebuilt mode by default
* fresh config loading in dev
* auto build in test when there is no metadata


idoros (approved):
Please check this works on codux before merging

packages/test-kit/src/with-feature.ts
```diff
@@ -262,7 +261,7 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}): WithF
         slowMo,
         persist,
         takeScreenshotsOfFailed = true,
-        buildFlow = process.env.WITH_FEATURE_BUILD_FLOW,
+        buildFlow = 'prebuild', //process.env.WITH_FEATURE_BUILD_FLOW,
```
idoros, line 264: remove comment

idoros (approved):

packages/engine-cli/README.md
```diff
@@ -1 +1,66 @@
...
+    publicPath,
+    buildTargets,
+    feature,
+    config,
+    run,
+    writeMetadataFiles,
+    publicConfigsRoute,
+  });
+}
+```
```
idoros, line 66:
```suggestion
```md
| Option                | Type            | Default Value | Description                                                                                       |
|-----------------------|-----------------|---------------|---------------------------------------------------------------------------------------------------|
| `buildTargets`        | `'node'`, `'web'`, `'both'` | `'both'`      | Target platforms for the build. Can be one of 'node', 'web', or 'both'.                          |
| `help`                | boolean         | `false`       | Displays help information if true.                                                               |
| `clean`               | boolean         | `true`        | Whether to clean the build directory before building.                                            |
| `watch`               | boolean         | `false`       | Rebuilds on file changes if true.                                                                |
| `dev`                 | boolean         | `watch` value | Enables development mode. Defaults to the value of `watch`.                                      |
| `run`                 | boolean         | `dev` value   | Runs the built application if true. Defaults to the value of `dev`.                              |
| `verbose`             | boolean         | `false`       | Enables verbose logging if true.                                                                 |
| `writeMetadataFiles`  | boolean         | `true`        | Whether to write metadata files during the build process.                                        |
| `runtimeArgs`         | JSON object     | `{}`          | Arbitrary arguments to pass at runtime, specified as a JSON string.                              |
| `feature`             | string          |               | Specifies a particular feature to build or run.                                                  |
| `config`              | string          |               | Path to a specific configuration file to use.                                                    |
| `publicPath`          | string          | `''` (empty string) | Base path for serving static files. Defaults to an empty string.                                 |
| `engineConfigFilePath`| string          |               | Path to the engine configuration file.                                                           |
| `publicConfigsRoute`  | string          | `'configs'`   | Route under which public configurations are served. Defaults to 'configs'.                       |

```

idoros (approved):

PR #2329, opened 2024-02-12
fix: better dispose of node env manager

barak007:
* remove usage of safe disposals for now.
* wait in all requests to current environment dispose 
* TODO: timeout


packages/runtime-node/src/metrics-utils.ts
```diff
@@ -2,15 +2,14 @@ import type { PerformanceMetrics } from './types';
 import { parentPort } from 'node:worker_threads';
 import type { ChildProcess } from 'node:child_process';
 import { Worker } from '@wixc3/isomorphic-worker/worker';
+import { isValidRpcMessage, isValidRpcResponse, rpcCall, getNextMessageId } from './micro-rpc';
 
-export function bindMetricsListener(
-    customFetcher: () => Promise<PerformanceMetrics> | PerformanceMetrics = localPerformanceFetcher,
-) {
+export function bindUniversalListener<T>(type: string, customFetcher: () => Promise<T> | T) {
```
idoros, line 7: move to rpc generic module

packages/runtime-node/src/metrics-utils.ts
```diff
@@ -2,15 +2,14 @@ import type { PerformanceMetrics } from './types';
...
     const handler = async (message: unknown) => {
-        if (isValidGetMetricsMessage(message)) {
+        if (isValidRpcMessage(message) && message.type === type) {
             const outgoingMessage = {
                 id: message.id,
-                metrics: await customFetcher(),
+                value: await customFetcher(),
```
idoros, line 12: Handle errors/timeout

idoros (changes requested):

packages/runtime-node/src/metrics-utils.ts
barak007: moved

idoros (approved):

PR #2337, opened 2024-02-18
feat: config manager to speedup config generation

barak007:
Loading config is slow. we need to eval the config. in dev time this causes a significant delay when refreshing the app. this PR introduce a config manager that can build an optimized version on the config and then load it and watch it.


packages/engine-cli/src/node-config-manager.ts
```diff
@@ -0,0 +1,140 @@
...
+    currentValue: Promise<unknown[]> | unknown[];
+    dispose(): Promise<void>;
+    build?: ReturnType<typeof deferred<unknown[]>>;
+};
+
+export class NodeConfigManager {
+    constructor(private config: esbuild.BuildOptions) {}
+    runningBuilds = new Map<string, BuildStats>();
+    hashConfig(entryPoints: string[]) {
+        return hash(entryPoints.join(','));
```
idoros, line 16: sort?

packages/engine-cli/src/node-config-manager.ts
```diff
@@ -0,0 +1,140 @@
...
+
+            const ctx = await this.createBuildTask(entryPoints, {
+                onStart() {
+                    const newBuild = deferred<unknown[]>();
+                    buildStats.error = undefined;
+                    buildStats.currentValue = newBuild.promise;
+                    buildStats.build = newBuild;
+                },
+                onEnd(files) {
+                    buildStats.error = undefined;
```
idoros, line 40: is there a case that build has an error and we still get to build end?

packages/engine-cli/src/node-config-manager.ts
```diff
@@ -0,0 +1,140 @@
...
+    }
+    async dispose() {
+        await this.disposeAll();
+    }
+    private async createBuildTask(
+        entryPoints: string[],
+        hooks: {
+            onStart(): void;
+            onEnd(files: undefined[]): void;
+            onError(err: Error): void;
```
idoros, line 84: connect `onError`

packages/engine-cli/src/node-config-manager.ts
```diff
@@ -0,0 +1,140 @@
...
+            format: 'cjs',
+            bundle: true,
+            write: false,
+        });
+
+        await Promise.all([ctx.watch(), deferredStart.promise]);
+        return ctx;
+    }
+}
+function deferred<T>() {
```
idoros, line 128: bring from `promise-assist`

packages/engine-cli/src/node-config-manager.ts
```diff
@@ -0,0 +1,140 @@
...
+        entryPoints: string[],
+        hooks: {
+            onStart(): void;
+            onEnd(files: undefined[]): void;
+            onError(err: Error): void;
+        },
+    ) {
+        const deferredStart = deferred<void>();
+        const ctx = await esbuild.context({
+            ...this.config,
```
idoros, line 89: maybe have tree-shaking by default

packages/engine-cli/src/node-config-manager.ts
barak007: nope. I don't think it's needed in this case. we pass a list of same config from the build so no variance.

packages/engine-cli/src/node-config-manager.ts
barak007: activated

packages/engine-cli/src/node-config-manager.ts
barak007: done

packages/engine-cli/src/node-config-manager.ts
barak007: connected

packages/engine-cli/src/node-config-manager.ts
barak007: nope.

packages/engine-cli/src/node-config-manager.ts
barak007: no sorting the order can will actually matter. but not for us.

idoros (approved):

PR #2348, opened 2024-02-21
fix: slow debug

barak007:
So when using esm bundle the devtools work very hard in order to show the `Module` keys in the scopes panel. using iife fixes that and we don't acctualy use the esbuild entry bundle as importable files (library)


packages/engine-cli/src/create-environments-build-configuration.ts
```diff
@@ -40,7 +40,7 @@ export function createBuildConfiguration(options: CreateBuildConfigOptions) {
     const commonConfig = {
         target: 'es2020',
         bundle: true,
-        format: 'esm',
+        format: 'iife',
```
idoros, line 43: add a comment so we can track why we picked `iife` over `esm`

idoros (approved):

PR #2415, opened 2024-03-21
feat: proper engine cli

barak007:
New cli with proper help.
Changed default for `forceAnalyze` to true


AviVahl (approved):
Looking good.

PR #2437, opened 2024-04-10
feat: allow engine config to define cli flags

barak007:
This PR allows the engine config to declare additional cli runtimeArgs embedded in the engine command.
This is not ideal because it can collide but I have a check to prevent it. maybe we can prefix them...but not for now.
Also added config and feature aliases.


AviVahl (approved):
Add pr description.
You've also added aliases.

PR #2440, opened 2024-04-11
feat(test-kit): introduce fixtures experimental dependencies installation

plisovyi:
By introducing new ENV variable `NEW_FIXTURE_DEPENDENCIES_INSTALLATION`, which makes getLoadedFeature retry installation twice with a 1s delay and 60s timeout


daomry:
IMO, the env variable is not a great idea, I'd just call it the new behaviour and be done with it.
The problem is that can be caused is that the env variable will create 2 different flows and make it hard to analyze if you're not aware of its value or existance

packages/test-kit/src/with-feature.ts
```diff
@@ -443,18 +444,32 @@ export function withFeature(withFeatureOptions: IWithFeatureOptions = {}): WithF
...
-                            shell: true,
-                            stdio: debugMode ? 'inherit' : 'ignore',
-                        });
+                        } satisfies SpawnSyncOptions;
+                        const shouldRetryInstall = !!process.env.NEW_FIXTURE_DEPENDENCIES_INSTALLATION;
+
+                        if (shouldRetryInstall) {
+                            const installTimeout = 10_000;
+
+                            await retry(
```
daomry, line 458: I think the test time might needs to be adjusted when successful
see adjustCurrentTestTimeout in core3utils/testing

daomry (changes requested):

plisovyi:
> The problem is that can be caused is that the env variable will create 2 different flows and make it hard to analyze if you're not aware of its value or existence
My idea was to put it under variable, use it in Codux, see if it solves our problems, and then use it by default.
> Is 10s enough for all tests?
TBH, I'm not sure; what do you think it should be? And yeah, test timeout should be adjusted I think

daomry (approved):

PR #2465, opened 2024-05-02
fix: forward respect pendingEnvs

barak007:
In the case of forward message between source env (processing), forwarder env (manager) and target env (main).
if the manager has the main environment as pending it will now wait until the target will be resolved to send any messages.

* loosen the type for `handleReady` to allow manual management  

TODO: 
- [x] test

### DIFF
--- a/packages/core/src/com/communication.ts
+++ b/packages/core/src/com/communication.ts
@@ -608,7 +608,10 @@ export class Communication {
         } else if (this.DEBUG) {
             console.debug(FORWARDING_MESSAGE(cleanMessageForLog(message), this.rootEnvId, env.id));
         }
-
+        if (this.pendingEnvs.get(env.id)) {
+            this.pendingMessages.add(env.id, () => this.post(this.resolveMessageTarget(env.id), message));
+            return;
+        }
         this.post(env.host, message);
     }
 
--- a/packages/core/src/com/communication.ts
+++ b/packages/core/src/com/communication.ts
@@ -35,7 +35,6 @@ import type {
     EventMessage,
     ListenMessage,
     Message,
-    ReadyMessage,
     StatusMessage,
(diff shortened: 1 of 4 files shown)


packages/core/src/com/communication.ts
```diff
@@ -765,7 +767,7 @@ export class Communication {
         }
     }
 
-    public handleReady({ from }: ReadyMessage): void {
+    public handleReady({ from }: { from: string }): void {
```
idoros, line 770: What flow would you like to call this manually? 
You can also remove all the `as ReadyMessage` from the other places that call this

idoros (approved):
Add a test

packages/core/src/com/communication.ts
barak007: manually state the env is ready or pending. (like the places you saw the cast at)
I removed all type casts

packages/core/test/node/com.spec.ts
```diff
@@ -159,6 +158,46 @@ describe('Communication', () => {
...
+
+        const mainCom = new Communication(middlemanHost, 'middle');
+        const bCom = new Communication(bHost, 'bEnv');
+        const aCom = new Communication(aHost, 'aEnv');
+
+        disposables.add(mainCom);
+        disposables.add(bCom);
+        disposables.add(aCom);
+
+        aCom.registerEnv('middle', middlemanHost);
```
idoros, line 174: Either register the middle to b as well or remove it from a to be persistent and less confusing

idoros (approved):
`envReady` and `handleReady` are horrible names for what they do

barak007:
> `envReady` and `handleReady` are horrible names for what they do

yes. when you look at them from the outside.
they was not designed to be public api. we need a new api for pending envs.

PR #2466, opened 2024-05-05
feat: enable static build

barak007:
This PR address missing `staticBuild` option in the new esbuild flow.

When building targets that requires fully static build like electron we need to build the config files to disk. 
This PR enables this feature in the esbuild flow.

This PR also fixes a build in the config loader when the `outDir` is absolute.

### DIFF
--- a/packages/engine-cli/src/top-level-config-plugin-esbuild.ts
+++ b/packages/engine-cli/src/top-level-config-plugin-esbuild.ts
@@ -1,6 +1,6 @@
 import { BuildOptions, Plugin } from 'esbuild';
 import fs from 'node:fs';
-import { dirname, join } from 'node:path';
+import { dirname, isAbsolute, join } from 'node:path';
 
 export interface PartialWebpackLoaderContext {
     query: string;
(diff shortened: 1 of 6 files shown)


packages/engine-cli/src/top-level-config-plugin-esbuild.ts
```diff
@@ -86,5 +86,6 @@ function emitConfigFile(
     const content = JSON.stringify(imported.default ?? imported);
     const configFileName = envName ? `${fileName!}.${envName}` : fileName;
     const configPath = `configs/${configFileName!}.json`;
-    filesToEmit.set(configPath, { content, path: join(rootContext, outDir, configPath) });
+    const path = isAbsolute(outDir) ? join(outDir, configPath) : join(rootContext, outDir, configPath);
```
AviVahl, line 89: use path.resolve in such cases, instead of checking for absolute yourself:
`path.resolve(rootContext, outDir, configPath)`

AviVahl (approved):
see small comment

PR #2476, opened 2024-05-22
fix(communication): don't alter error messages in cross-environment calls

alisey:
These messages are displayed to users, so it's not ok to append debug information to them.

Also it violates developer's expectation to throw an error on one end and receive the same error on the other end.


idoros (changes requested):
Good catch, please fix the [test](https://github.com/wixplosives/engine/blob/7ce1b6bd553305bc23606951cb69bc83588aeacd/packages/core/test/communication.spec.ts#L88) so this requirement won't rely on a comment (just need to replace the matcher to equal)

idoros (changes requested):
I recall why we added this feature. It was crucial for debugging certain cases in Codux. Since this is still relevant for debugging, can you please preserve the message modification under the `this.DEBUG` flag. This is a manual mode that a developer need to enable, and it doesn't run during tests or in production.

idoros (approved):

PR #2477, opened 2024-05-23
fix(communication): ensure that errors can be sent via WebSocket

alisey:
In communication, we have a custom function `serializeError()` that converts an error to a plain object before sending it between environments. I've noticed that this function sanitizes the error object to prevent exceptions when sending it via postMessage, but there are a few minor issues:

1. It uses `JSON.parse(JSON.stringify(...))` which can itself throw an exception, e.g. due to circular references, and that exception is not handled. The cross-environment method call will simply not receive a response.
2. The comment in the function claims that `JSON.parse(JSON.stringify(...))` is necessary to ensure that the error can go through `postMessage`. If that was the case, a simple `structuredClone()` would suffice. In reality, the error also needs to survive sending through WebSocket.
3. The function sanitizes some properties but not others. Even though it's highly unlikely that `error.stack` will be anything other than a string, it wouldn't hurt to make sure of that.

I've fixed those issues, added support for [error.cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) and renamed the function from `serializeError()` to `errorToJson()`.


idoros (approved):
Did you check codux doesn't rely on the deleted `serializeError`? 
It's a shame all of our packages export everything using `*` and everything is public api making refactors harder.

alisey:
> Did you check codux doesn't rely on the deleted serializeError?

Yes, it's not used in Codux.

> It's a shame all of our packages export everything using * and everything is public api making refactors harder.

Yes. Also adds a bunch of unnecessary code to bundles.

PR #2500, opened 2024-06-20
feat(core): allow passing entry options via script query

zetmate:
For the abomination. Providing entry options by other means than the search params will allow to keep the iframe url clean which is vital when running the full app instead of a board.

### DIFF
--- a/packages/core/src/helpers/web.ts
+++ b/packages/core/src/helpers/web.ts
@@ -13,10 +13,10 @@ export function injectScript(win: Window, rootComId: string, scriptUrl: string) 
 }
 
 interface EngineWebEntryGlobalObj {
-    document?: Document,
-    location?: Location
+    document?: { currentScript: { dataset: { engineRunOptions?: string | null } } | null },
+    location?: { search?: string }
 
-    engineEntryOptions(options: { urlParams: URLSearchParams; envName: string }): IRunOptions;
+    engineEntryOptions?: (options: { currentRunOptions: IRunOptions; envName: string }) => IRunOptions;
 }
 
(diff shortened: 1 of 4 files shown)


packages/core/src/helpers/web.ts
```diff
@@ -9,3 +9,22 @@ export function injectScript(win: Window, rootComId: string, scriptUrl: string)
         win.document.head.appendChild(scriptEl);
     });
 }
+
+export function getEngineEntryOptions(envName: string) {
```
barak007, line 13: I know it wasn't there before but can you add a simple unit test to this function it has like 3 layer fallback that I would like to make sure is in tact.

my suggestion is to pass the `globalThis` with a narrow type of all the required fields for this function to work.

Also it's not safe to split `url` string by `?` you could use `URL.parse` and get the `.searchParams` from there.

at the end this function can return a regular `Map` and not `URLSearchParams` I did that because I only had that in hand already. it will make the final loop cleaner by checking the same map keys no additional set is needed

packages/core/src/runtime-configurations.ts
```diff
@@ -1,11 +1,5 @@
 import { IRunOptions, type TopLevelConfig } from './types.js';
 
-// eslint-disable-next-line @typescript-eslint/no-namespace
-declare namespace globalThis {
```
barak007, line 4: This was a tricky thing to do...I didn't want the cast `(globalThis as unknown as Window)` in isMainEntrypoint because it's not true this code can also run in node.

barak007 (approved):

PR #2501, opened 2024-06-21
feat(dashboard): distinguish parent and child relations on a graph

VladyslavG:
I had a task to refactor some fixtures in the Codux. I decided to use our feature graph for that and it was really useful.
However when viewing feature graph I noticed that all links - parent and child are highlighted in the same color. I could not understand looking at a graph where was my feature requested as a child. Take a look at this:

image

I hovered over file-assets and it is impossible to tell which related feature uses file-assets vs which is used by file-assets.

So I propose to use different colors for parent links vs child links. I used this approach personally in my abovementioned task and it helped me a lot! So I decided to create this PR, maybe it will help other people when playing with features

After applying my changes:
image

You can clearly see different colored lines and distinguish parent vs child relations.
If you don't like the suggested color, please feel free to fix it.

### DIFF
--- a/packages/dashboard/src/components/feature-graph.tsx
+++ b/packages/dashboard/src/components/feature-graph.tsx
@@ -85,8 +85,8 @@ export const FeatureGraph = memo(({ selectedFeatureGraph }: IFeatureGraphProps) 
                         return linkedNodes[0] === graphNode
                             ? linkedNodes[linkedNodes.length - 1]
                             : linkedNodes[linkedNodes.length - 1] === graphNode
-                            ? linkedNodes[0]
-                            : 0;
+                              ? linkedNodes[0]
+                              : 0;
                     })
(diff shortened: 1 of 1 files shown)


barak007:
have you tried to run `yarn engine analyze` ? 
I think we should invest more there...maybe even move this graph to it. 
my plan is to remove the original dashboard package in the feature.

barak007:
@VladyslavG what was the motivation here? do you need better analysis?

VladyslavG:
> yarn engine analyze

my motivation was to understand relationships between features visually without reading the code for each feature. And that graph (with my tweak) was very helpful.
`yarn engine analyze` doesn't help me to visualize feature dependencies

VladyslavG:
> the entire old dashboard got removed.

so there is now no way to visualize feature dependencies?

PR #2503, opened 2024-06-26
chore: remove peer Dependence on webpack in engine-scripts.

barak007:
We are in the process of removing webpack and peerDependencies on them are forces install on consumer side. 
* all users of engine removed webpack dependencies.


AviVahl:
ah, don't forget to run `npm i` to update lock file.

PR #2526, opened 2024-07-14
feat: use node's native import() to load modules

AviVahl:
target is now node16, which means we still transpile our code to cjs (no "type": "module" yet).

we now retain the dynamic import() calls as-is, rather than transpiling them to require() calls.

this will make it easier for user projects to move to native esm.


barak007:
BTW
I don't like the way you squash locally and force push. 
it makes all the comment notifications (and their comments) 404, and also I don't see what you are struggling with.

AviVahl:
This didn't land as using dynamic import requires registration of a loader if you run directory from sources.
We do such thing downstream, so hold off until we have [loaders](https://nodejs.org/api/esm.html#loaders) in place.

PR #2536, opened 2024-07-29
feat(engine)!: remove `electron-host` and `electron-commons` packages

plisovyi:
As Electron support in the Engine does not require Engine tooling anymore.


AviVahl:
avoid regenerating the package-lock.json when local `node_modules` already exist. do a fresh `git clean -fdx`.
otherwise, it can remove all urls from lockfile , like it did in this case.

plisovyi:
> avoid regenerating the package-lock.json when local `node_modules` already exist. do a fresh `git clean -fdx`. otherwise, it can remove all urls from lockfile

image

🙇🏼

AviVahl (approved):

PR #2537, opened 2024-07-30
feat(engine-cli): move `generate` command from deprecated engineer to engine-cli

plisovyi:
As `engineer` is deprecated, and scripts are centralized in `engine-cli`, moving `engineer create` to `engine generate <featureName>` before `engineer` will be removed.

The method was deleted in the engineer (the package should go on its own).

Removed irrelevant documentation (there's CLI help now, screenshot:)
Screenshot 2024-07-30 at 17 43 28


packages/scripts/src/application/application.ts
```diff
@@ -189,33 +185,18 @@ export class Application {
...
-            fs,
-            featureName,
-            targetPath,
-            templatesDirPath,
-            featureDirNameTemplate,
-        });
+    /** @deprecated use engine-cli instead (`engine generate FeatureName`) */
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public async create(..._: unknown[]) {
+        throw new Error('Deprecated');
```
AviVahl, line 199: remove this completely. we wanted to remove this package and no longer use its cli. no point in placeholder error messages

plisovyi: OK, I will just remove it here.
There's also a possibility to remove `electron-commons` and `electron-host` I think; it can be done as a one major, maybe 🤔

cylon3035 (approved):

PR #2539, opened 2024-08-03
infra: migrate to eslint@9

AviVahl:


AviVahl:
@barak007 newer (rc version) `eslint-plugin-react-hooks` mis-detects `Feature.use` as bad react hook usage. It supports react 19, where `use` is now a thing.

I turned off `react-hooks/rules-of-hooks` for now, but we may want to rename this fn to avoid hitting this lint rule's false positive.

PR #2546, opened 2024-08-12
fix(test-kit): tests screenshots path is broken on windows

### LINKED ISSUE #2545: [CI issue]: Uploading tests screenshots doesn't work for windows

zetmate:
Fixes #2545 
Fixes the logic of sanitising the screenshots path. 
The current logic replaces every ':' with a '-' which on windows leads to a path 'C-\Users\...' instead of 'C:\Users\...'. 
This PR adds a lookbehind to the regex so it skips the ':' if it goes after a capital letter which is a first symbol in the path


barak007 (approved):

zetmate:
This is actually a bad fix, it should also make sure that the `^[A-Z]:` is followed by a slash. Otherwise it allows : with whatever, just preceded with a capital letter. I'll update.

PR #2555, opened 2024-08-25
feat: added engine shutdown signal to features setup

barak007:
This is a first step for better handling of engine disposal. 
now it is expected from features to use `engineShutdownSignal` instead of `onDispose` hook for **cancelable** control flows.
notice that handling abort event should not throw and it's possible to register a `onDispose` hook inside.


packages/core/test/node/runtime.spec.ts
```diff
@@ -707,6 +707,30 @@ describe('feature disposal', () => {
...
+            id = 'test' as const;
+            api = {};
+        }
+        const onEvent = spy((msg) => msg);
+        entryFeature.setup(mainEnv, ({ engineShutdownSignal, onDispose }, {}) => {
+            onDispose(() => onEvent('dispose'));
+            engineShutdownSignal.addEventListener('abort', () => onEvent('abort'));
+        });
+
+        const engine = await runEngine({
```
idoros, line 724: consider adding a test to check that abort can be called before engine is ready

idoros (approved):

PR #2557, opened 2024-09-01
refactor(communication): better errors

plisovyi:
This PR aims to refactor/clean up Communications regarding debug logging and error handling in preparation for better processing of such errors during error monitoring. Errors would have proper names and messages attached to them with redacted information. 

* In Communication feature setup removed `CommunicationOptions` as they are inferred all way back to class property definition
* Inlined `RemoteAPIServicesMapping` type and moved `ConfigEnvironmentRecord` to [packages/core/src/com/types.ts](https://github.com/wixplosives/engine/pull/2557/files#diff-b214f55abd9ad6b5d8131b548fd20652b60e31ccede4c85ae7e7c9b6d7e6307eL119-R129)
* Inlined logs from [packages/core/src/com/logs.ts](https://github.com/wixplosives/engine/pull/2557/files#diff-60434e2c7b8580fc90e0a0fbdb34ea4e6a92bba051bddbb9d345551f99cc77cf) to `communication.ts` as messages make more sense near the action they describe and are not reused in multiple places; used `this.log?.()` pattern to make sure log code doesn't run if debug is off
* Unified communication errors in [packages/core/src/com/communication-errors.ts](https://github.com/wixplosives/engine/pull/2557/files#diff-3f83c8f3a24c70f3da99ce4a04ff66fef8f4143e8d85371a1949b54ca253d80e)
* In [packages/core/src/com/communication.ts](https://github.com/wixplosives/engine/pull/2557/files#diff-41b934e9c12240c900ceb249d4129d8cb603b86657e672115851d437684512c1) simplified some typing and improved naming + moved methods not dependent on the instance to helpers; switched some methods with lots of parameters to accept an object and other stylistic non-functional changes; also, `callMethod` method made private


packages/core/src/com/communication.ts
```diff
@@ -1,182 +1,182 @@
...
+        publicPath: '',
+        connectedEnvironments: {},
+    };
+    private disposing = false;
+
     constructor(
-        private host: Target,
-        id: string,
+        private readonly host: Target,
+        private readonly id: string,
```
barak007, line 89: did you check that no one is using this `id` outside?

barak007, line 89: this is so confusing that we still have `rootEnvId` and this `id` now. choose one.

barak007: I cannot follow if this change is valid because of that

packages/core/src/com/communication.ts
```diff
@@ -1,182 +1,182 @@
...
     ) {
-        this.options = {
-            warnOnSlow: this.DEBUG,
-            publicPath: '',
-            connectedEnvironments: {},
-            ...options,
-        };
-        this.rootEnvId = id;
-        this.rootEnvName = id.split('/')[0]!;
+        this.options = { ...this.options, ...options };
```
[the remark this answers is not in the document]
plisovyi: You are right; would `Object.assign(this.options, options)` be better handled in this case? 🤔

packages/core/src/com/communication.ts
```diff
@@ -1,182 +1,182 @@
...
      */
     public registerAPI<T extends {}>({ id }: IDTag, api: T): T {
-        if (!this.apis[id]) {
-            this.apis[id] = api;
-            this.applyApiDirectives(id, api);
-            return api;
-        } else {
-            throw new Error(DUPLICATE_REGISTER(id, 'RemoteService'));
-        }
+        if (this.apis[id]) throw new DuplicateRegistrationError(id, 'RemoteService');
```
barak007, line 147: add `{}` to the if

packages/core/src/com/communication.ts
```diff
@@ -1,182 +1,182 @@
...
-        this.topology[envName] = envUrl;
+    public getPublicPath(): string {
+        return this.options.publicPath;
     }
 
-    public subscribeToEnvironmentDispose(handler: (envId: string) => void) {
+    public subscribeToEnvironmentDispose(handler: (envId: string) => void): void {
         this.disposeListeners.add(handler);
     }
-    public unsubscribeToEnvironmentDispose(handler: (envId: string) => void) {
```
barak007, line 177: where is this function?

packages/core/src/com/communication.ts
```diff
@@ -340,46 +347,53 @@ export class Communication {
...
-            pending.push(promise);
+            });
+            statuses.push(promise);
         }
-        return Promise.allSettled(pending).then((results) => {
-            const statuses: Record<string, ReturnType<Communication['getComStatus']> | PromiseRejectedResult> = {};
+
+        return Promise.allSettled(statuses).then((results) => {
+            const statuses: Record<string, Status | PromiseRejectedResult> = {
+                [this.id]: this.getStatus(),
```
barak007, line 392: This moves the time of the snapshot to after we take all other statuses. I'm not sure that it's matters. but does it?

packages/core/src/com/communication.ts
```diff
@@ -654,81 +672,89 @@ export class Communication {
...
+                });
+                resolve();
             } else {
-                res();
+                resolve();
             }
         } else {
-            if (serviceComConfig[method]?.listener) {
-                const handlersBucket = this.handlers.get(this.getHandlerId(envId, api, method));
+            if (!serviceComConfig[method]?.listener) throw new UnConfiguredMethodError(api, method);
```
barak007: add `{}` to if

packages/core/src/com/communication.ts
plisovyi: OK, `rootEnvId` gone in favor of just `id`

packages/core/src/com/communication.ts
```diff
@@ -737,275 +763,225 @@ export class Communication {
...
-        if (env.id !== this.rootEnvId) {
-            return host;
-        } else {
-            if (host instanceof BaseHost) {
-                return host.parent || host;
-            }
-            return this.getPostEndpoint(host);
-        }
+        const env = this.environments[envId] ?? this.environments[this.id];
+        if (!env) return HOST_REMOVED;
```
barak007, line 771: add `{}` to if

packages/core/src/com/communication.ts
plisovyi: 👍🏼 Yeah, got cleaned up as unused, but it was wrong on my part; I checked that it is unused both internally and externally, but externally, it should be used alongside `subscribeToEnvironmentDispose` I think

barak007 (commented):
This is a huge PR! that is really hard to track the changes in. and it's in the most used place in our system. it feels like it should come out as Major change since even small timing of events can change the system behavior. 

It looks fine on the surface but hard to tell.

also stop with if's with no `{}`

if you wanted to improve errors it could have been mush smaller PR.

packages/core/src/com/communication.ts
plisovyi: I wouldn't think it would... get status would be invoked on line 392 when creating an object on 391. before it did happen on 357, but between that and now only things that are happening are registering status calls to others; not waiting to get all those statuses. 

But it is indeed change in how things work, so I will revert to how it was. Great catch thank you!

plisovyi:
> This is a huge PR! that is really hard to track the changes in.
> if you wanted to improve errors it could have been mush smaller PR.

Well, you are right. It is a big one, and just improving errors could be extracted into a much smaller PR, which I did here —https://github.com/wixplosives/engine/pull/2568; so please, let's review that one instead.

Next time, let's not even get into reviewing "huge PRs" if you think there's too much in the way of reviewing. If it is possible, it is not too hard to decouple only required changes into their own PR. And while indeed this is the very crucial core part, I think we should encourage changes and evolution by making sure there are no hiccups in the process; changes in 2568 could have been extracted 2 weeks ago and there'd be no need in this whole back and forth.

PR #2567, opened 2024-09-12
fix: race condition when engine was not ready and terminate was called.

barak007:
In tests it's possible that the termination request was fired before main entry point was finished.
we used to only register to the termination event after it was finished and we missed the terminate event.  

There is no test in this PR since there is a tech-debt to allow for such tests: https://github.com/wixplosives/engine/issues/2569


barak007:
https://github.com/wixplosives/engine/issues/2569

idoros (approved):

packages/scripts/src/create-node-entrypoint.ts
```diff
@@ -73,13 +73,27 @@ import { workerData } from "node:worker_threads";
...
+        console.log('[${env.name}]: Termination Requested. Waiting for engine.');
+    }
+    unbindTerminationListener();
+    unbindMetricsListener();
+    try {
+        const engine = await running;
+        console.log('[${env.name}]: Terminating');
+        return engine.shutdown();
+    } catch (e) {
+        return;
```
AviVahl, line 93: would this not swallow shutdown errors we would like to see?

PR #2595, opened 2024-11-18
fix: handle error being thrown in shutdown of run

matangeorgi:
when the `catch` scope is being triggered in the `run` function, the catch will execute:
```
await this.shutdown();
```

If shutdown fails for some reason we have no information of what happened.
Adding an extra catch is for adding the missing information.


AviVahl (commented):
Test please

AviVahl (approved):

PR #2610, opened 2024-12-08
feat!: fully native esm

AviVahl:
- all packages are now "type": "module"
- replaced commonjs calls with their esm counterparts
- ensure all dynamic imports now use pathToFileURL to properly work on win32
- removed now-unneeded helper that bypassed typescript transpilation to get to the "real" import()
- changed default output format to esm


packages/runtime-node/src/node-env-manager.ts
```diff
@@ -252,7 +252,7 @@ function connectWorkerToHost(envName: string, worker: ReturnType<typeof runWorke
             worker.postMessage(message.data);
         };
         const handleInitializeError = (e: AnyMessage) => {
-            rej(new Error(`failed initializing environment ${envName} with error message: ${JSON.stringify(e.data)}`));
+            rej(new Error(`failed initializing environment ${envName}`, { cause: e.data }));
```
AviVahl, line 255: this one actually prints the error now :)

packages/engine-cli/src/engine-build.ts
```diff
@@ -87,7 +87,7 @@ export async function runEngine({
     outputPath = fs.resolve(rootDir, outputPath);
 
     const jsOutExtension = '.js' as '.js' | '.mjs';
-    const nodeFormat = jsOutExtension === '.mjs' ? 'esm' : 'cjs';
+    const nodeFormat = 'esm' as 'esm' | 'cjs';
```
AviVahl, line 90: ^ this

packages/engine-cli/src/node-config-manager.ts
```diff
@@ -130,6 +131,7 @@ export class NodeConfigManager {
                             try {
                                 const text = outputFiles[0]!.text;
                                 const module = { exports: { default: undefined as any } };
+                                const require = createRequire(import.meta.url);
```
AviVahl, line 134: not sure about this. we're in esm context now. manual evaluation with require?

PR #2618, opened 2024-12-16
feat(iframe): get run option from parent

cylon3035:
now, it will be possible to add to the iframe
```
<script data-engine-run-options="fetch-options-from-parent=true" src="./preview.web.js"></script>
```
or start the iframe with src of `./preview.web.js?fetch-options-from-parent=true`
then use `installRunOptionsInitMessageHandler` on the side of the parent window to have a working iframe env.

It will be used to decouple id generation from iframe dom node creation.

Note: iframe initialization in the engine since we have a [tech debt here](https://github.com/wixplosives/codux/issues/24381). We have an initializer directly in codux and we need to move it back to make sense.


plisovyi (approved):

cylon3035:
@barak007, after discussing with @plisovyi we can argue that:
* timeout does not contribute to a better user experience. it is the responsibility of the parent to timeout if init is stuck.
* time out reject will interfere with debug breakpoints
therefore I will remove the timeout part

PR #2619, opened 2024-12-17
feat(core): support feature optional dependencies

idoros:
This PR introduces an `optionalDependencies` configuration option for defining features. 

Optional dependencies will only be initialized and made available if another feature includes them in its (required) `dependencies` definition. This ensures that a feature can leverage another feature dynamically, based on its availability in the current running "flavor."

### DIFF
--- a/packages/core/src/runtime-engine.ts
+++ b/packages/core/src/runtime-engine.ts
@@ -75,9 +75,9 @@ export class RuntimeEngine<ENV extends AnyEnvironment = AnyEnvironment> {
         const toProcess = [...entryDeps];
         while (toProcess.length) {
             const feature = toProcess.shift()!;
-            if (this.allRequiredFeatures.has(feature.id)) continue;
             this.allRequiredFeatures.add(feature.id);
             for (const dep of feature.dependencies()) {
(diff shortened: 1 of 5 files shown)


packages/core/src/runtime-engine.ts
```diff
@@ -68,6 +70,19 @@ export class RuntimeEngine<ENV extends AnyEnvironment = AnyEnvironment> {
...
     }
 
+    private preRun(entryDeps: FeatureClass[]) {
+        // populate required features
+        const toProcess = [...entryDeps];
+        while (toProcess.length) {
+            const feature = toProcess.shift()!;
+            if (this.allRequiredFeatures.has(feature.id)) continue;
+            this.allRequiredFeatures.add(feature.id);
+            for (const dep of feature.dependencies()) {
```
barak007, line 80: move the check to the loop `if (this.allRequiredFeatures.has(feature.id)) continue;`
it will be much better

barak007: because most of the features are coming from deps and not top level.

barak007 (approved):

PR #2621, opened 2024-12-18
feat(iframe init): change init handlers api

cylon3035:
Add the ability to post back run options after receiving the initial message.


plisovyi (approved):

idoros:
@cylon3035 Was there an issue adding a test for this case?

PR #2655, opened 2025-02-19
feat: enhance dashboard server launch with correct entry point logging

barak007:
This fix the issue of always showing main.html env in log


idoros (approved):
We need to have tests here

PR #2664, opened 2025-03-06
feat: add support for custom pre-build configurations 

matangeorgi:
Add the option to pass to the engine config an array of paths which should be bundled before the main bundle.

example of usage:
```ts
customPreBuildDist: [
        './src/code-editor/vendors/monaco-css-worker.ts',
        './src/code-editor/vendors/monaco-generic-worker.ts',
        './src/code-editor/vendors/monaco-html-worker.ts',
        './src/code-editor/vendors/monaco-json-worker.ts',
    ],
```

### DIFF
--- a/packages/engine-cli/src/engine-build.ts
+++ b/packages/engine-cli/src/engine-build.ts
@@ -17,7 +17,7 @@ import { createBuildConfiguration } from './create-environments-build-configurat
 import { readEntryPoints, writeEntryPoints } from './entrypoint-files.js';
 import { EntryPoints, EntryPointsPaths } from './create-entrypoints.js';
 import { resolveRuntimeOptions } from './resolve-runtime-options.js';
-import type { EngineConfig, PreBuildConfig } from './types.js';
+import type { EngineConfig } from './types.js';
 import { analyzeFeatures } from './find-features/analyze-features.js';
 import { ENGINE_CONFIG_FILE_NAME } from './find-features/build-constants.js';
 import { runPreBuilds } from './pre-build.js';
@@ -45,7 +45,7 @@ export interface RunEngineOptions {
     staticBuild?: boolean;
     customEntrypoints?: string;
     title?: string;
-    customPreBuildDist?: PreBuildConfig[];
+    customPreBuildDist?: string[];
 }
 
 export async function runEngine({
--- a/packages/engine-cli/src/engine-build.ts
+++ b/packages/engine-cli/src/engine-build.ts
@@ -295,11 +295,11 @@ export async function runEngine({
     };
 }
 
-function shouldBuildWeb(buildTargets: 'node' | 'web' | 'both' | 'electron') {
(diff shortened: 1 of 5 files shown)


packages/engine-cli/src/pre-build.ts
```diff
@@ -0,0 +1,52 @@
...
+    buildConfiguration: BuildConfiguration,
+    buildTargets: 'node' | 'web' | 'both' | 'electron',
+    verbose = false,
+): Promise<void> {
+    if (!preBuildsPaths.length) {
+        return;
+    }
+
+    if (verbose) {
+        console.log(`Running ${preBuildsPaths.length} pre-builds before main build...`);
```
idoros, line 18: If there is a pre-build configuration, then a minimal log should be given even without `verbose`

packages/engine-cli/src/index.ts
```diff
@@ -13,3 +13,4 @@ export { runLocalNodeManager } from './run-local-mode-manager.js';
 export * from './types.js';
 export type { IExecutableApplication, RunningFeature } from './types.js';
 export { checkWatchSignal } from './watch-signal.js';
+export { runPreBuilds } from './pre-build.js';
```
idoros, line 16: Is there a reason to expose this as API from the index?

packages/engine-cli/src/engine-build.ts
```diff
@@ -198,6 +201,10 @@ export async function runEngine({
         favicon,
     });
 
+    if (customPreBuildDist) {
+        await runPreBuilds(rootDir, customPreBuildDist, buildConfigurations, buildTargets, verbose);
+    }
+
     if (watch) {
```
idoros, line 208: Add a notice to the PR description that pre-builds do not support watch mode and are built only at the beginning.  If there is a documentation somewhere then that should be added there aswell. Maybe you can add it as part of the log when they start running

idoros (changes requested):
I believe the CLI flag might be missing in this implementation. To ensure everything works as expected, I’d suggest adding a test that checks the API from an external perspective, rather than focusing on a unit test for the internal logic.

packages/engine-cli/src/pre-build.ts
matangeorgi: done

packages/engine-cli/src/engine-build.ts
matangeorgi: added this log (even with verbose set as false):
```
console.log('Notice that pre-builds does not support watch mode');
```

matangeorgi:
> I believe the CLI flag might be missing in this implementation. To ensure everything works as expected, I’d suggest adding a test that checks the API from an external perspective, rather than focusing on a unit test for the internal logic.

This is not a CLI flag, it is being passed as a field in the engine config, i have checked it using copypack in codux to see that it works.
if we still want to add a test to check it in integration, which function should we test, engine()? runEngine()?

matangeorgi:
Closing this pr since it is possible to achieve through esbuild-plugins.

PR #2712, opened 2025-05-07
feat(node-env-manager): add naive health check

cylon3035:
add  GET /health - 200 after all feature environments in node are run


0xRezaa (approved):
lgtm

PR #2731, opened 2025-07-01
refactor: remove config api from node env manager - seems not use

alissawix:
remove config api from node env manager app - seems not use

move createFeatureEnvironmentsMapping to cli (where it's used)

move files and imports accordingly

all code is copy pasted


packages/runtime-node/src/node-env-manager.ts
```diff
@@ -60,29 +50,10 @@ export class NodeEnvManager implements IDisposable {
         process.env.ENGINE_FLOW_V2_DIST_URL = this.importMeta.url;
         const disposeMetricsListener = bindMetricsListener(() => this.collectMetricsFromAllOpenEnvironments());
         const verbose = Boolean(runtimeOptions.get('verbose'));
-        const topLevelConfigInject = parseInjectRuntimeConfigConfig(runtimeOptions);
 
         const staticDirPath = fileURLToPath(new URL('../web', this.importMeta.url));
         const { port, socketServer, app, close } = await launchEngineHttpServer({ staticDirPath, ...serverOptions });
         runtimeOptions.set('devServerPort', port.toString());
-        app.get<{ configName: string[] }>('/configs/*configName', async (req, res) => {
```
AviVahl, line 68: lets talk about this

PR #2851, opened 2026-04-24
feat(com): propagate caller identity across cross-environment calls and subscriptions

vladbazl:
Problem.
Server-side environments need to know who is making API calls. 
Identity is derived from the Socket.IO handshake (e.g. cookies, auth headers). When a request is forwarded through multiple envs — client → processing → workspace — every server-side handler in the chain must see the original client's identity.

Solution.
1. `autoLaunch` receives `identityExtractor`. Example:
```
await manager.autoLaunch(new Map([['feature', 'caller-identity']]), {
    identityExtractor: (handshake) => ({ userId: handshake.auth?.userId ?? 'anonymous' }),
});
```
2. In Node env entrypoint configures `callerContext` in `COM` feature config with using a standard [AsyncLocalStorage](https://nodejs.org/api/async_context.html#asynclocalstoragerunstore-callback-args) library. Example:
```
new RuntimeEngine(
        env,
        [
            ...(workerData
                ? [
                      COM.configure({
                          config: {
                              host: new ParentPortHost(env.env),
                              id: env.env,
                              callerContext: new AsyncLocalStorage(),
                          },
                      }),
                  ]
                : []),
            ...topLevelConfig,
        ],
        new Map(options?.entries() ?? []),
    ).run(Feature);
```
3. Registered APIs (services) can use a `getCurrentCaller()` in any method. Example:
```
import { getCurrentCaller } from '@dazl/engine-core';
...
const userIdentity = getCurrentCaller();
```


packages/core/src/com/caller-context.ts
```diff
@@ -0,0 +1,34 @@
...
+        return true;
+    } catch {
+        callerStore = null;
+        return false;
+    }
+}
+
+export function getCurrentCaller(): unknown {
+    return callerStore?.getStore();
+}
```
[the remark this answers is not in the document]
vladbazl: updated types

packages/runtime-node/src/ws-node-host.ts
```diff
@@ -98,6 +110,9 @@ export class WsServerHost extends BaseHost implements IDisposable {
         const existingSocket = this.clientIdToSocket.get(clientId);
 
         this.clientIdToSocket.set(clientId, socket);
+        if (this.identityExtractor) {
+            this.identityStore.set(clientId, this.identityExtractor(socket.handshake));
```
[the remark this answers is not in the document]
vladbazl: added error handling

packages/core/src/com/communication.ts
```diff
@@ -984,7 +993,9 @@ export class Communication {
     private async handleCall(message: CallMessage): Promise<void> {
         try {
             const args = deserializeApiCallArguments(message.data.args);
-            const data = await this.apiCall(message.origin, message.data.api, message.data.method, args);
+            const data = await runWithCaller(this.isServer, message.callerIdentity, () =>
+                this.apiCall(message.origin, message.data.api, message.data.method, args),
+            );
```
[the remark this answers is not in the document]
vladbazl: It is for instances created in 'com' feature, not for 'clients-host-com'. 'clients-host-com' receives identity from 'WsHost' in incoming message.

packages/core/src/com/caller-context.ts
```diff
@@ -0,0 +1,34 @@
...
+let callerStore: CallerContext | undefined | null;
+
+async function initCallerContext() {
+    if (callerStore !== undefined) {
+        return callerStore !== null;
+    }
+    try {
+        const moduleName = 'node:async_hooks';
+        const { AsyncLocalStorage } = await import(moduleName);
```
AviVahl: @engine/core is isomorphic.
please do not use node APIs in isomorphic code.
avoid using a dynamic import in a try-catch to account for this.

packages/core/src/com/caller-context.ts
```diff
@@ -0,0 +1,34 @@
+interface CallerContext {
+    getStore<T = unknown>(): T | undefined;
+    run<R>(identity: unknown, callback: () => R): R;
+}
+let callerStore: CallerContext | undefined | null;
```
AviVahl: avoid module state

vladbazl: why?

AviVahl: Because it leaks between different runs in the same runtime environment, as it depends on module evaluation to initialize and cleanup. This fact makes harder to manage and not testing friendly, by design.

It is a hack/cheat to use it a in a controlled environment, where you'd like to remove leftovers and ensure cleanup.

Not the first instance I'm commenting about this. Told the AI Squad the same when they did it in one of the tools.

packages/core/src/com/caller-context.ts
vladbazl: refactoring this part

vladbazl: Made it a configuration of COM feature [optional], without a conditional dependency (dynamic import) on NodeJS lib.

packages/core/src/com/caller-context.ts
vladbazl: This part was updated, now it is configurable through COM feature settings and each thread can receive a different implementation. 
The `CallerContext` is still set in module though, this is needed to make the Caller Context API nice and avoid changes (e.g. dependencies on Communication instance) in every place where caller identity is needed.
Also this functionality is covered with tests.

PR #2858, opened 2026-05-22
feat: add identity to socket and message

alissaVrk:
the goal is to add identity of original caller to message.

- user passes `identityExtractor` to the engine on setup, used in the `WsServerHost` upon a new socket connection to map socket to identity.
- add `callerIdentity` on each incoming message to the `WsServerHost`
- in communication, `forwardMessage` we just forward the identity.  

for client => env A **=>** env B case, we want to pass client identity to env B. since engine doesn't know which socket/clientId is calling the `callMethod`.
- user passes `getCallerIdentity` to communication. 
- in communication `callMethod`, we add  `callerIdentity` to the message based on `getCallerIdentity`

so the responsibility to have the  `callerIdentity` available during the call is on the engine user.


packages/core/src/com/communication.ts
idoros, line 1: I don't think we need to attach `callerIdentity` on callbacks aswell for the flow of, right?
envA -> envB.service.method() -> return value (the context here still has the initial caller)

Another communication case is even emitter, when an event is emitted, the receiver won't have a caller identity. I believe that's fine, but just so we know this is intentional.

packages/runtime-node/src/ws-node-host.ts
```diff
@@ -99,6 +110,16 @@ export class WsServerHost extends BaseHost implements IDisposable {
 
         this.clientIdToSocket.set(clientId, socket);
 
+        const extractor = this.identityExtractor;
+        if (extractor) {
+            this.callWithErrorHandling(() => {
+                const identity = extractor(socket.handshake, clientId);
+                if (identity) {
+                    this.identityStore.set(clientId, identity);
+                }
```
idoros, line 121: Clear client identity connection if no identity is resolved. Lets assume someone can reconnect with the same client is but with unknown auth somehow and take actions as the auth user

packages/runtime-node/src/ws-node-host.ts
```diff
@@ -128,6 +149,7 @@ export class WsServerHost extends BaseHost implements IDisposable {
             // modify message to be able to forward it
             message.from = fromId;
             message.origin = originId;
+            message.callerIdentity = this.identityStore.get(clientId);
```
idoros, line 155: Lets add a comment here - we always set the callerIdentity, even as `undefined` - not allowing the client to somehow send a `callerIdentity` on the message and bypassing the identity mechanism

idoros (changes requested):

packages/core/src/com/communication.ts
alissaVrk: right

alissaVrk: will add to event message, you are right

idoros (approved):

PR #2859, opened 2026-05-25
add wrapper function for api calls

alissaVrk:
add ability to pass a wrapper function to api calls.
can be used to get identity from the message.


packages/core/src/com/communication.ts
```diff
@@ -946,7 +950,9 @@ export class Communication {
         const dispatcher = this.eventDispatchers.get(namespacedHandlerId)?.dispatcher;
         if (dispatcher) {
             this.eventDispatchers.delete(namespacedHandlerId);
-            const data = await this.apiCall(message.origin, message.data.api, message.data.method, [dispatcher]);
+            const data = await this.options.apiCallWrapper(message, () =>
+                this.apiCall(message.origin, message.data.api, message.data.method, [dispatcher]),
+            );
```
idoros, line 955: previously the code explicitly awaited the api call. here the wrapper might ignore it.

packages/core/src/com/communication.ts
```diff
@@ -991,7 +999,9 @@ export class Communication {
     private async handleCall(message: CallMessage): Promise<void> {
```
idoros, line 999: Why do we need to wrap in other places that are not the main entry for message (this one)

packages/core/src/communication.feature.ts
```diff
@@ -10,7 +10,7 @@ import { Slot } from './entities/slot.js';
 import { RUN_OPTIONS, ENGINE } from './symbols.js';
 import { type LoggerTransport, LogLevel } from './types.js';
 import { WindowInitializerService } from './com/window-initializer-service.js';
-import { CallerIdentity } from './com/message-types.js';
+import { CallerIdentity, Message } from './com/message-types.js';
```
idoros: type import

packages/core/test/node/com.spec.ts
idoros, line 1: Not sure what changed in this file 😖

idoros: I refactored to minimize the changes: https://github.com/dazl-dev/engine/pull/2859/commits/6de05c7f62107efe3b5fc9970e9fc46293cb9c00

alissaVrk: indentation, wrapped tests in describes.
it was too annoying

idoros (approved):

